PR #6, opened 2026-04-20
[proxy]: Add Admin/Workflow Service proxies

pseudomuto:
Introduces the internal/proxy package, which provides gRPC proxy implementations for AdminService and WorkflowService. Each proxy forwards every RPC to an upstream client using forwardUnary (for unary calls) and forwardBidiStream (for the AdminService bidi streaming RPC), preserving incoming gRPC metadata on all calls.

Will be building on this in subsequent PRs to handle things like stripping auth headers/metadata, etc.


internal/proxy/adminservice.go
```diff
@@ -0,0 +1,194 @@
+package proxy
+
+import (
+	"context"
+
+	adminservicev1 "go.temporal.io/server/api/adminservice/v1"
+)
+
+type AdminServiceProxy struct {
```
liam-lowe, line 9: I'm probably re-igniting a conversation that has been asked before we both joined. 

How do we keep these proxies in sync - when we go to do a release - should we fail release if we aren't covering all endpoints of specified temporal server version for that service?

This will be more relevant (and maybe not a problem) depending on the slice of use cases this proxy supports - ie S2S proxy supports a subset (and is typically ACL'ed by customer to allow less than that subset).

liam-lowe (approved):
LGTM!

internal/proxy/adminservice.go
pseudomuto: That's a great point, and an open question, at least for v1. I would imagine, we'll want a CI/CD check to ensure, or at least notify us, when we're out of date. I also assume that the ACL idea will at some point be borrowed here, since likely that exists for a reason

PR #8, opened 2026-04-20
[config]: Add Temporal cluster type and restructure config types

pseudomuto:
- Add Temporal cluster type for proxying to a real Temporal server via plain gRPC
- Replace LocalCluster (Inbound/Outbound) with a flat Listener field — ClusterType
  already determines the connection role, so a single listen address suffices
- Replace RemoteCluster with Upstream (address + pool size); promote Name and Type
  to top-level Cluster fields
- Rename RemoteType → ClusterType
- Require cluster Type in validation; omitting it is now an error

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -9,6 +9,120 @@ import (
 	"github.com/temporalio/temporal-proxy/internal/config"
 )
 
+func TestLoad_ClusterTypeValidation(t *testing.T) {
+	t.Parallel()
+
+	t.Run("missing cluster type returns error", func(t *testing.T) {
+		t.Parallel()
+
+		yaml := `
+clusters:
+  - listener:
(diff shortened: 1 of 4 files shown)


internal/config/config_test.go
```diff
@@ -9,6 +9,120 @@ import (
...
+	t.Parallel()
+
+	t.Run("missing cluster type returns error", func(t *testing.T) {
+		t.Parallel()
+
+		yaml := `
+clusters:
+  - listener:
+      hostPort: :8233
+`
```
liam-lowe, line 22: formatting 😭

liam-lowe (approved):

internal/config/config_test.go
pseudomuto: Which part? The blank line at the end?

PR #9, opened 2026-04-20
[creds]: Add transport/creds package for handling (m)TLS

pseudomuto:
Introduces `internal/transport/creds` with three gRPC credential implementations, each exposing a consistent two-method interface:

  - DialOption() grpc.DialOption  — configures outbound (client) connections
  - ServerOption() grpc.ServerOption — configures inbound (server) connections

Credential types:

  Insecure — disables transport security entirely; suitable for local
  development or service-mesh environments where encryption is handled
  elsewhere.

  TLS — one-way TLS; the server presents a cert/key pair and the client
  verifies against the system root CA pool. No client certificate required.

  MTLS — mutual TLS; both sides present certificates and verify against a
  shared CA. Supports optional ServerName override and InsecureSkipVerify
  (testing only). Enforces TLS 1.2 minimum and restricts server cipher
  suites to AES-256-GCM and AES-128-GCM.

The `Credentials` interface is now defined in `internal/server` and the package default is updated to use `creds.NewInsecure()`. The previous `internal/server/credentials.go` is removed as it is superseded by this package.


liam-lowe (approved):
elegant!

internal/transport/creds/mtls.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+// DialOption returns a [grpc.DialOption] that configures the outbound
+// connection with mutual TLS. The client presents its certificate/key pair and
+// verifies the server against the CA pool loaded from caFile.
+func (c *MTLS) DialOption() (grpc.DialOption, error) {
+	cert, err := tls.LoadX509KeyPair(c.certFile, c.keyFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load server key pair: %w", err)
```
liam-lowe: client key pair?

internal/transport/creds/tls.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// NewTLS returns a [TLS] credential that loads the server certificate and
+// private key from certFile and keyFile respectively.
+func NewTLS(certFile, keyFile string) *TLS {
+	return &TLS{
+		certFile: certFile,
+		keyFile:  keyFile,
+	}
+}
```
liam-lowe: Might be misleading for client - cert/key are args but not required? Maybe split into NewClientTLS/NewServerTLS.

internal/transport/creds/tls.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load server key pair: %w", err)
+	}
+
+	cfg := &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   minTLSVersion,
+	}
+
+	return grpc.Creds(credentials.NewTLS(cfg)), nil
```
liam-lowe, line 74: This doesn't set same ciphersuites as mtls. For my own understanding - what's the reasoning?

internal/transport/creds/mtls.go
```diff
@@ -0,0 +1,109 @@
...
+		// certificate hostname. Useful when the server's certificate CN does not
+		// match its dial address.
+		ServerName string
+	}
+)
+
+// NewMutualTLS returns an [MTLS] credential that loads the CA certificate from
+// caFile and the client certificate/key pair from certFile and keyFile. opts
+// provides additional configuration; a zero-value [MTLSOptions] is valid.
+func NewMutualTLS(caFile, certFile, keyFile string, opts MTLSOptions) *MTLS {
```
liam-lowe: nit: `NewMutualTLS` returns `MTLS`

Would prefer to keep same semantic across both `MTLS` or `MutualTLS`

liam-lowe (approved):

internal/transport/creds/tls.go
pseudomuto: Yeah, it's a little weird to pass "" for both for clients. I'll update.

internal/transport/creds/tls.go
pseudomuto: This was an oversight, I'll update both to use the same. Nice catch!

**Reasoning for Specifying Ciphers**

TLS 1.2 includes a suite of ciphers that are considered less secure. By specifying the particular ciphers in mtls, we're saying you can use tls1.2 (old) if you want, but only these ciphers which support AES-GCM (considered more secure than block chaining) and ECDHE for key exchange.

In Go, tls1.3 and above this setting is ignored, as any cipher considered insecure is removed via their patch system when they build releases.

https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/crypto/tls/common.go;l=724;drc=9f30ca9189831b8061a6ee46c0a2a331b4864559

internal/transport/creds/tls.go
pseudomuto: Took your suggestion and made two distinct functions

internal/transport/creds/tls.go
pseudomuto: Made them both require those ciphers for tls1.2, and added some comments about why.

PR #11, opened 2026-04-22
[crypto]: Introduce crypto package for encryption

pseudomuto:
Adds the internal/crypto package, which implements envelope encryption
for Temporal payload data.

In envelope encryption, each payload is encrypted with a short-lived
Data Encryption Key (DEK). The DEK itself is then encrypted by a
customer-managed Key Encryption Key (KEK) — such as an AWS KMS or GCP
Cloud KMS key — and stored alongside the payload as DEKMaterial.

Types:

- DEK: a 256-bit AES-256-GCM key. Ciphertext is prefixed with a
  randomly generated nonce to prevent replay attacks.
- DEKMaterial: the data needed to later recover a DEK — the KEK ID and
  the base64-encoded encrypted DEK bytes.
- KEK: an interface for customer-managed Key Encryption Keys, expected
  to be backed by an external KMS (AWS KMS, GCP Cloud KMS, etc.).
- KEKRegistry: a thread-safe registry mapping Temporal namespaces to
  KEKs. Supports a default fallback key and per-namespace overrides.
  An fx.Module is provided for dependency injection.

Encryption flow:
1. Call NewDEK() to generate a fresh 256-bit random DEK.
2. Call DEK.Encrypt(ctx, plaintext) to encrypt the payload.
3. Call KEKRegistry.Encrypt(ctx, namespace, dek) to wrap the DEK,
   producing a DEKMaterial.
4. Store the DEKMaterial alongside the ciphertext in payload metadata.

Decryption flow:
1. Extract DEKMaterial from payload metadata.
2. Call KEKRegistry.Decrypt(ctx, material) to unwrap the DEK.
3. Call DEK.Decrypt(ctx, ciphertext) to recover the plaintext.

Direct support for AWS/GCP/static KEK implementations is not included
here and will be added in subsequent PRs.


internal/crypto/kms.go
```diff
@@ -0,0 +1,111 @@
...
+
+	if len(dek) != keyBytes {
+		return DEK{}, fmt.Errorf("invalid DEK for KEK: %s", m.KEKID)
+	}
+
+	return DEK(dek), nil
+}
+
+// Close closes all registered KEKs and removes them from the registry. Returns
+// the first error encountered; any remaining KEKs are not closed in that case.
```
liam-lowe: > Returns the first error encountered

Any reason not to try close everything?

```go
  var errs []error
  for k, kek := range r.keks {
      if err := kek.Close(); err != nil {
          errs = append(errs, fmt.Errorf("close %s: %w", k, err))
      }
      delete(r.keks, k)
  }
```

internal/crypto/kms.go
```diff
@@ -0,0 +1,111 @@
...
+func (r *KEKRegistry) Close() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
```
liam-lowe: Lock held during close - which will incur network IO (kms, et al ) ?

What happens to other threads waiting for lock after `Close()` completes? I imagine there are interleaving race conditions where the lock is released by encrypt/decrypt - then `close()` is called - but a closed key can still be used (depends on kms source as to whether this is a concern / what error would be returned on use). Ie the lock protected the map, but we still have references to the values in the map.

Is a "safer" (maybe just easier to reason about)  approach to implement a restriction on map + slow drain of active usage, before closing keys:
- introduce a registry `closed` which is checked in each function under lock (to prevent register/encrypt/decrypt when registry is (or is trying to) close. 
- add a waitgroup to track in-flight usages (semaphore incremented under encrypt/decrypt lock acquisition,  decremented after k use.
- upon close:
  - acquire lock briefly to only mark the registry closed - don't acquire lock yet for closing keys. Don't let    new acquirers of the lock increment the semaphore.
  - await the waitgroup to be fully released
  - then when certain nobody else is manipulating registry - can close the keys.



There's probably other approaches, but underlying concerns are:

- acquiring lock to do network IO
- lock protects map, but not usage of the values in the map

internal/crypto/dek.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+)
+
+// NewDEK generates a new random 256-bit Data Encryption Key.
+func NewDEK() DEK {
+	k := make([]byte, keyBytes)
+	if _, err := rand.Read(k); err != nil {
+		// Only fails on legacy Linux systems (kernel < 3.17) without /dev/urandom.
+		// See https://github.com/golang/go/blob/master/src/crypto/rand/rand.go
+		panic(fmt.Sprintf("crypto/rand unavailable: %v", err))
```
liam-lowe: personally, would prefer to return the error. (or validate this earlier in startup chain and swallow here as unreachable).

internal/crypto/dek.go
```diff
@@ -0,0 +1,82 @@
...
+// Decrypt decrypts the ciphertext ct using AES-256-GCM. The ciphertext must be
+// prefixed with the nonce, as produced by [DEK.Encrypt].
+func (d DEK) Decrypt(ctx context.Context, ct []byte) ([]byte, error) {
+	gcm, err := d.newGCM()
+	if err != nil {
+		return nil, err
+	}
+
+	ns := gcm.NonceSize()
+	pt, err := gcm.Open(nil, ct[:ns], ct[ns:], nil)
```
liam-lowe: Do we need to guard against the slice oob `(ct[:ns], ct[ns:])` if ciphertext is too short?

internal/crypto/dek.go
```diff
@@ -0,0 +1,82 @@
...
+	return pt, nil
+}
+
+func (d DEK) newGCM() (cipher.AEAD, error) {
+	block, err := aes.NewCipher(d[:])
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cipher block: %w", err)
+	}
+
+	gcm, err := cipher.NewGCM(block)
```
liam-lowe, line 73: > Generate a fresh DEK (256-bit AES-GCM key) per payload or batch.

What are your thoughts around what defines a batch? Seems like this is called per `Encrypt()` (one DEK per payload)?

internal/crypto/kms.go
```diff
@@ -0,0 +1,111 @@
...
+	return &KEKRegistry{
+		keks: keys,
+	}
+}
+
+// Register adds or replaces the KEK with the given id in the registry.
+func (r *KEKRegistry) Register(id string, k KEK) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.keks[id] = k
```
liam-lowe: What if `id` already exists? Do we need to close the existing key first? Do we need to drain existing usage?

internal/crypto/kms.go
```diff
@@ -0,0 +1,111 @@
...
+	KEKRegistry struct {
+		mu   sync.RWMutex
+		keks map[string]KEK
+	}
+)
+
+// NewKEKRegistry returns a KEKRegistry pre-populated with the provided keys,
+// where each map key is the ID used to look up the corresponding KEK.
+func NewKEKRegistry(keys map[string]KEK) *KEKRegistry {
+	return &KEKRegistry{
```
liam-lowe: stores keys by reference, not by copy?

original caller can technically mutate without going through registry:

one approach would be to create `keks` via copy of `keys` to separate the two. 

But personally it could be easier to maintain / reason about with single entrance for `Register`. Why does a newRegistry need to initialize keys - can the caller just loop through and call `Register` themselves if they need this? If it's really a required usage pattern - can add `RegisterBatch` to reduce bloat of lock acquisition.

internal/crypto/dek.go
pseudomuto: That makes sense, we can panic in an `init` func if necessary and swallow here

internal/crypto/dek.go
pseudomuto: It depends on how defensive you think we should be. If this is being used (it's internal) to decrypt only ciphertexts that were encrypted with a DEK, this will never be a problem. If we expect people might take arbitrary bytes and attempt to decrypt them with a DEK, then it'd go 💥.

internal/crypto/dek.go
pseudomuto: Yes, I'm getting ahead of myself here I think. The idea was that we may have a number of commobpb.Payload (what we're ultimately encrypting) come in in a single request. In that case, we could avoid new DEKs for each one, and use a single, per-request DEK. However, I haven't added the API for that, so I can either:

* Update `Encrypt` to take `pts ...[]byte`
* Remove the batch idea until it actually comes up (leaning towards YAGNI)

liam-lowe: Happy to call YAGNI for now! Was just understanding potential intent! Thanks!

pseudomuto:
@liam-lowe I reworked this significantly to address some of your feedback. Some things worth calling out in particular:

* No more weirdness around rand.Read errors. I simply returned it from the constructor. I didn't see an upside to swallowing/panicking
* Close returns all errors and it idempotent (subsequent calls will not re-close)
* There is still a lock in close (via Once.Do), but I've left a note there about why I think it's ok here.

internal/crypto/kek.go
```diff
@@ -0,0 +1,149 @@
...
+func WithKeyForNamespace(ns string, k KEK) KEKRegistryOption {
+	return kekRegOpt(func(r *KEKRegistry) {
+		r.keks[ns] = k
+	})
+}
+
+// Encrypt encrypts the given DEK using the KEK registered for the specified namespace.
+// It returns DEKMaterial containing the KEK ID and the base64-encoded ciphertext.
+func (r *KEKRegistry) Encrypt(ctx context.Context, ns string, dek *DEK) (*DEKMaterial, error) {
+	k, ok := r.keks[ns]
```
liam-lowe, line 91: should this use default key as fallback ?

internal/crypto/fx.go
```diff
@@ -0,0 +1,30 @@
...
+	},
+))
+
+// KEKRegistryParams holds the fx-injected parameters for constructing a KEKRegistry.
+type KEKRegistryParams struct {
+	fx.In
+
+	// Default key to use when namespace override doesn't exist.
+	// When nil, a blank KEK is used (no encryption occurs).
+	DefaultKey KEK `optional:"true"`
```
liam-lowe, line 26: optional `true` - but it's wired in at line 10. Would this throw error if not passed?

internal/crypto/kek.go
```diff
@@ -0,0 +1,149 @@
...
+func WithDefaultKey(k KEK) KEKRegistryOption {
+	return kekRegOpt(func(r *KEKRegistry) {
+		r.defaultKey = k
+	})
+}
+
+// WithKeyForNamespace registers k for ns, used when encrypting or decrypting DEKs for that namespace.
+func WithKeyForNamespace(ns string, k KEK) KEKRegistryOption {
+	return kekRegOpt(func(r *KEKRegistry) {
+		r.keks[ns] = k
```
liam-lowe, line 84: Given KEK is only on `New` should we throw exception on duplicate key - no overrides allowed. Could prevent a failure mode in future.

internal/crypto/dek.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to decrypt ciphertext: %w", err)
+	}
+
+	return pt, nil
+}
+
+func (d *DEK) getOrCreateCipher() (cipher.AEAD, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
liam-lowe: Is this lock here too aggressive: we only write once, but read many times.

The read doesn't need a lock (the value can't change when set). 

1. My first thought: let's move the lock below the next clause

```golang
if d.gcm != nil {
		return d.gcm, nil
	}
```

But this would lead to parallelism race conditions (subsequent `getOrCreate()` while first is still running leads to multiple re-creates.)

We could do this defensively:
```golang
	if d.gcm != nil {
		return d.gcm, nil
	}

	d.mu.Lock()
	defer d.mu.Unlock()

	if d.gcm != nil {
		return d.gcm, nil
	}

    // create lock
```

This way we only acquire lock on initial write (and any parallel reads during the write). Maybe there's a more idiomatic go way of solving this. idempotent create of immutable value.

2. My second thought is:

Does this matter. How often will a DEK be re-used. If it's only used serially - who cares. But I imagine this is critcial path across and shared state

internal/crypto/kek.go
```diff
@@ -0,0 +1,149 @@
...
+	})
+
+	return r.closeErr
+}
+
+func (f kekRegOpt) apply(r *KEKRegistry) {
+	f(r)
+}
+
+func (k *nilKEK) ID() string                                           { return "EMPTY_DEK" }
```
liam-lowe: nit: `EMPTY_DEK`

liam-lowe (approved):

internal/crypto/fx.go
pseudomuto: I don't believe so. Marking it as optional, means `nil` (zero-val) is cool in this case. On line 10, we would call `WithDefaultKey(nil)` which is not an issue. This did cause me to realize I have an error in the constructor though (thank you!).

internal/crypto/kek.go
pseudomuto: Yes, I'll add a test and ensure the fallback works

internal/crypto/kek.go
pseudomuto: That's probably a good idea. As YAML files get big, I could see a duplicate sneaking in there. NIce catch! I'll update.

internal/crypto/dek.go
pseudomuto: #12 is where DEKs become longer-lived (still short-term), so I think re-use is common. I'll update the logic to use the if/lock/if you proposed.

internal/crypto/kek.go
pseudomuto: What am I missing? 😄

internal/crypto/dek.go
pseudomuto: I actually think the initial `d.gcm != nil` check needs to have a lock (could be RLock). I'm going to remove this issue by using sync.Once to initialize the cipher and be done with it

internal/crypto/kek.go
liam-lowe: sorry: this is file for KEK -> but the string returns "EMPTY_DEK"

internal/crypto/kek.go
pseudomuto: Updated to use the default key

internal/crypto/fx.go
pseudomuto: Fixed the case where passing nil would lead to an NPE

internal/crypto/dek.go
pseudomuto: Ended up dropping locks/sync.Once in favour of simply creating the gcm in the constructor.

internal/crypto/kek.go
pseudomuto: Resolved by panicking when receiving a duplicate. Normally, I wouldn't use panic to handle this, but I think it makes sense in this case. This is initialization logic, and there's nothing to be done to make it right (update your config file appropriately is the course of action.).

PR #12, opened 2026-04-28
[crypto]: Add Sealer to handle envelope encryption

pseudomuto:
Adding Sealer as the primary entry point for envelope encryption. It manages per-namespace DEK rotation and exposes two operations:

- Seal(ctx, id, plaintext) encrypts data under the current DEK for the given ID (typically a Temporal namespace) and returns an Envelope containing the AES-256-GCM ciphertext and the DEKMaterial needed to recover the DEK later.

- Open(ctx, envelope) recovers the DEK from the KEKRegistry using the embedded DEKMaterial, then decrypts and returns the original plaintext.

DEKs will be pre-rotated before expiry via a periodic `Refresh` call, keeping key generation off the Seal hot path. The rotation window is controlled per-ID through KeyConfig (Lifetime and ExpirationBuffer). If Refresh hasn't run by the time a key expires, Seal rotates inline as a fallback.

### DIFF
--- a/internal/crypto/sealer.go
+++ b/internal/crypto/sealer.go
@@ -1,0 +1,235 @@
+package crypto
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+)
+
+type (
+	// Sealer manages per-ID DEK rotation and envelope encryption.
+	//
+	// Each ID (typically a Temporal namespace) has its own [DEK] with an
+	// independent rotation schedule defined by a [KeyConfig]. DEKs are
+	// pre-rotated by a background [Sealer.Refresh] call before their
+	// expiration buffer is reached, keeping key generation off the hot path.
+	//
+	// Use [Sealer.Seal] to encrypt data and [Sealer.Open] to decrypt it.
+	Sealer struct {
+		mu       sync.RWMutex
+		keys     map[string]*sealerDEK
+		registry *KEKRegistry
+		now      func() time.Time
+	}
+
+	sealerOptions struct {
+		config map[string]*KeyConfig
+		now    func() time.Time
+	}
+
+	// SealerOption configures a [Sealer] during construction.
+	SealerOption func(*sealerOptions)
+
+	// Envelope is the output of a [Sealer.Seal] call. It carries the
+	// AES-256-GCM ciphertext together with the [DEKMaterial] required to
+	// recover the DEK for decryption via [Sealer.Open].
+	Envelope struct {
+		CipherText  []byte
+		KeyMaterial *DEKMaterial
+	}
+
+	// KeyConfig defines the config (e.g. rotation policy) for a single key slot.
+	KeyConfig struct {
+		Lifetime         time.Duration // How long a DEK remains valid
+		ExpirationBuffer time.Duration // How far before expiry [Sealer.Refresh] should pre-rotate the key.
+	}
+
+	sealerDEK struct {
+		key    *DEK             // The current DEK
+		exp    time.Time        // When this key expires
+		incr   time.Duration    // How long each DEK is valid for
+		expBuf time.Duration    // How long before expiry should we regenerate
+		now    func() time.Time // Clock source, shared with the parent Sealer
+	}
+)
+
+// NewSealer constructs a Sealer backed by r, applying opts in order.
+// It generates an initial [DEK] for every ID registered via [WithKeyConfig].
+// Returns an error if any DEK generation fails.
+func NewSealer(r *KEKRegistry, opts ...SealerOption) (*Sealer, error) {
+	sopts := &sealerOptions{
+		config: make(map[string]*KeyConfig),
+		now:    time.Now,
+	}
+
+	for _, opt := range opts {
+		opt(sopts)
+	}
+
+	s := &Sealer{
+		keys:     make(map[string]*sealerDEK),
+		registry: r,
+		now:      sopts.now,
+	}
+
+	for k, v := range sopts.config {
+		key, err := NewDEK()
+		if err != nil {
+			return nil, err
+		}
+
+		s.keys[k] = &sealerDEK{
+			incr:   v.Lifetime,
+			expBuf: v.ExpirationBuffer,
+			now:    s.now,
+		}
+
+		// NB: No locks needed in constructor
+		s.keys[k].rotate(key)
+	}
+
+	return s, nil
+}
+
+// WithKeyConfig registers a [KeyConfig] rotation policy for id.
+func WithKeyConfig(id string, cfg KeyConfig) SealerOption {
+	return func(s *sealerOptions) {
+		if _, ok := s.config[id]; ok {
+			panic(fmt.Sprintf("Duplicate key config for '%s'", id))
+		}
+
+		s.config[id] = &cfg
+	}
+}
+
+// WithNowFunc overrides the clock used for DEK expiry checks and rotation
+// timestamps. Intended for testing; production code should use the default
+// [time.Now].
+func WithNowFunc(fn func() time.Time) SealerOption {
+	return func(s *sealerOptions) {
+		s.now = fn
+	}
+}
+
+// Seal encrypts data using the DEK for id, returning an [Envelope] that
+// contains the ciphertext and the [DEKMaterial] needed to decrypt it later.
+func (s *Sealer) Seal(ctx context.Context, id string, data []byte) (*Envelope, error) {
+	dek, err := s.getOrRefreshKey(id)
+	if err != nil {
+		return nil, err
+	}
+
+	ct, err := dek.Encrypt(ctx, data)
+	if err != nil {
+		return nil, err
+	}
+
+	m, err := s.registry.Encrypt(ctx, id, dek)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Envelope{
+		CipherText:  ct,
+		KeyMaterial: m,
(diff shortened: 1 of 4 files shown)


internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Refresh rotates every expired DEK. It is intended to be called periodically
+// from a background goroutine so that key rotation stays off the [Sealer.Seal]
+// hot path.
+func (v *Sealer) Refresh() error {
+	// Find expired keys without acquiring a write lock.
+	expKeys := make([]string, 0, len(v.keys))
+	v.mu.RLock()
+	for k, v := range v.keys {
```
liam-lowe: v is being overwritten in this scope.

Also a little misleading - `for k, v` implies for `key, value` - but then the "`value"` is an item in `v.keys` - which is a key. -- so the value `v` is of type `key` and if locally scoped over the `v *Sealer`. 

Personally - I'd rename `v` to `s`  - unsure what `v` is shorthand for.

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+	// pre-rotated by a background [Sealer.Refresh] call before their
+	// expiration buffer is reached, keeping key generation off the hot path.
+	//
+	// Use [Sealer.Seal] to encrypt data and [Sealer.Open] to decrypt it.
+	Sealer struct {
+		mu       sync.RWMutex
+		keys     map[string]*expiringDEK
+		config   map[string]*KeyConfig
+		registry *KEKRegistry
+		now      func() time.Time
```
liam-lowe, line 23: Massive nit - you can ignore - but I wrote this on first read:

---

Reading sequentially ( I haven't read the rest of PR yet), this is a little hard to read (I'm sure it will make sense with context) - my personal take is that the idiomatic go-ism of short-lengthed names is making this harder to read than necessary: 

- `keys` is a `namespace : expiringDEK` map. 
  - As a reader/user of struct - I expect this is actually a local "DEK with expiration". 
  - To me: an "`expiringDEK`" reads like it would be a heap.
  - "expiring" is a _current_ property of the DEK, which might not be the only truth as the module grows - I'm assuming it's just the Sealer-specific implementation of a DEK (which atm, only has an expiration over the base DEK)
- `KeyConfig` doesn't indicate which key it is config for. 

I think I prefer types that specific these are explicitly for Sealer DEK's:
- `expiringDEK` -> `SealerDEK`
- `KeyConfig` -> `SealerDEKConfig`

This then makes the opportunity for future abstraction cleaner (if `SealerDEK` and `SealerDEKConfig` are combined in a similar way to `KEKRegistry`

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+	KeyConfig struct {
+		Lifetime         time.Duration // How long a DEK remains valid
+		ExpirationBuffer time.Duration // How far before expiry [Sealer.Refresh] should pre-rotate the key.
+	}
+
+	expiringDEK struct {
+		key    *DEK             // The current DEK
+		exp    time.Time        // When this key expires
+		incr   time.Duration    // How long each DEK is valid for
+		expBuf time.Duration    // How long before expiry should we regenerate
```
liam-lowe, line 52: what is difference between :

- `KeyConfig.Lifetime` and `expiringDEK.incr`
- `KeyConfig.ExpirationBuffer` and `expiringDEK.expBuf`

1. Any reason for duplication
2. Any reason for inconsistent naming

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+// [time.Now].
+func WithNowFunc(fn func() time.Time) SealerOption {
+	return func(s *Sealer) {
+		s.now = fn
+	}
+}
+
+// Seal encrypts data using the DEK for id, returning an [Envelope] that
+// contains the ciphertext and the [DEKMaterial] needed to decrypt it later.
+func (v *Sealer) Seal(ctx context.Context, id string, data []byte) (*Envelope, error) {
```
liam-lowe: nit: why is `Sealer` shorthanded to `v`

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+	// happen when the ExpirationBuffer < the Refresh interval (e.g. Refresh every minute, but ExpirationBuffer is 30s),
+	// or with just bad luck/timing.
+	//
+	// The goal remains to refresh keys optimistically before they expire, so we don't generate keys in the hot path
+	// (when encrypting data).
+	v.mu.Lock()
+	defer v.mu.Unlock()
+
+	// Re-check after acquiring the write lock: Refresh (or another Seal) may
+	// have already rotated this key in the window between RUnlock and Lock.
```
liam-lowe, line 214: nice!

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	return s, nil
+}
+
+// WithKeyConfig registers a [KeyConfig] rotation policy for id.
+func WithKeyConfig(id string, cfg KeyConfig) SealerOption {
+	return func(s *Sealer) {
+		s.config[id] = &cfg
+	}
```
liam-lowe: do we need to check for non-nil/validity here . Should sealer enforce a minimum lifetime / expiration buffer.

internal/crypto/sealer.go
liam-lowe: Is the KeyConfig transient - does it need to exist on `Sealer` or should `KeyConfig` and `expiringDEK` be combined in a parent container.

internal/crypto/sealer.go
liam-lowe: should we panic on duplicates

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ct, err := dek.Encrypt(ctx, data)
+	if err != nil {
+		return nil, err
+	}
+
+	m, err := v.registry.Encrypt(ctx, id, dek)
```
liam-lowe: Do we cache key material - or do we make a call to KMS (et al) for each encryption call?

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+	expKeys := make([]string, 0, len(v.keys))
+	v.mu.RLock()
+	for k, v := range v.keys {
+		if v.isExpired() {
+			expKeys = append(expKeys, k)
+		}
+	}
+	v.mu.RUnlock()
+
+	// Generate DEKs without the lock as well.
```
liam-lowe, line 169: makes sense!

Calling out it introduces race condition - but I don't think there's a current concern:

- multiple concurrent `Refresh` - could lead to duplicate work. 
  - I don't think this is a concern as I imagine `Refresh` is called sequentially by a single go-routine. 
- could also happen with `getOrRefreshKey` refresh scenario as well

I don't think this is a concern as we would only be doing duplicate generation work in an edge-case scenario - but maybe future changes make this impact more meaningful

internal/crypto/sealer.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return pt, nil
+}
+
+// Refresh rotates every expired DEK. It is intended to be called periodically
+// from a background goroutine so that key rotation stays off the [Sealer.Seal]
+// hot path.
+func (v *Sealer) Refresh() error {
+	// Find expired keys without acquiring a write lock.
+	expKeys := make([]string, 0, len(v.keys))
```
liam-lowe: might as well do in the lock - incase we change `v.keys` outside of constructor in the future.

liam-lowe (approved):
LGTM!

Shared some nits & inline thoughts - but nothing blocking!

internal/crypto/sealer.go
pseudomuto: I originally named it "Vault", but didn't update the references. I'll switch it up...this is weird

internal/crypto/sealer.go
pseudomuto: At the moment we don't cache and call KMS each time to encrypt/decrypt the dek. I nice addition would be to cache locally until expiry, or some LRU. I like this idea, but would tackle it independently.

internal/crypto/sealer.go
pseudomuto: I renamed the receiver to s, I think k, v makes sense now

internal/crypto/sealer.go
pseudomuto: I renamed expiringDEK (a DEK that expires) to sealerDEK so it's more explicit in terms of who's using it. I didn't see any reason to export it.

internal/crypto/sealer.go
pseudomuto: KeyConfig is transient, I've introduced `sealerOptions` here, so we don't need it on Sealer itself. We just need it to be able to build the initial keys

internal/crypto/sealer.go
pseudomuto: I added a panic on duplicates block. This is in line with what was done for the KEKRegistry

internal/crypto/sealer.go
pseudomuto: I think this is ok, as you mentioned. It runs the risk of duplicate work, but not deadlock. Holding the lock to generate the keys could be worse since nothing can be done while in progress.

PR #14, opened 2026-05-01
[crypto]: Cache DEKMaterial to avoid KMS calls

pseudomuto:
Sealer.Seal called registry.Encrypt (a KMS network call) on every invocation, even when the active DEK had not changed. Sealer.Open called registry.Decrypt on every invocation, even for envelopes whose DEK had already been decrypted. At volume, this KMS round-trip dominates latency.

This commit introduces two caches:

Seal: sealerDEK now holds the KMS-wrapped DEKMaterial after the first wrap call. All subsequent Seal calls for the same DEK reuse the cached material, making exactly one KMS call per DEK rotation per namespace. The cache is cleared on rotation so stale material is never returned. A key-identity guard (sd.key == dek) under the write lock handles the race where Refresh rotates the DEK between the snapshot and the cache store.

Open: Sealer now has a bounded LRU cache keyed by the base64-encoded EncryptedDEK string. A cache hit returns the decrypted DEK directly, skipping KMS entirely. The cache size defaults to 100 entries and is configurable via WithCacheSize; passing 0 disables the cache.

> NB: 100 keys in the LRU cache is 300KB on the high end. Typically 1.5-2KB
per DEK.

### DIFF
--- a/internal/crypto/sealer.go
+++ b/internal/crypto/sealer.go
@@ -5,7 +5,12 @@ import (
 	"fmt"
 	"sync"
 	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+	"golang.org/x/sync/singleflight"
 )
+
+const defaultCacheSize = 100
 
 type (
 	// Sealer manages per-ID DEK rotation and envelope encryption.
@@ -17,15 +22,18 @@ type (
 	//
 	// Use [Sealer.Seal] to encrypt data and [Sealer.Open] to decrypt it.
 	Sealer struct {
-		mu       sync.RWMutex
-		keys     map[string]*sealerDEK
-		registry *KEKRegistry
-		now      func() time.Time
+		mu        sync.RWMutex
+		keys      map[string]*sealerDEK
+		registry  *KEKRegistry
+		now       func() time.Time
+		dekCache  *lru.Cache[string, *DEK] // keyed by EncryptedDEK; nil if cacheSize == 0
+		encryptor singleflight.Group       // coalesces concurrent registry.Encrypt calls per namespace
 	}
 
 	sealerOptions struct {
-		config map[string]*KeyConfig
-		now    func() time.Time
+		config    map[string]*KeyConfig
+		now       func() time.Time
+		cacheSize int // <=0 = no cache
 	}
 
 	// SealerOption configures a [Sealer] during construction.
@@ -46,11 +54,12 @@ type (
 	}
 
 	sealerDEK struct {
-		key    *DEK             // The current DEK
-		exp    time.Time        // When this key expires
-		incr   time.Duration    // How long each DEK is valid for
-		expBuf time.Duration    // How long before expiry should we regenerate
-		now    func() time.Time // Clock source, shared with the parent Sealer
+		key      *DEK             // The current DEK
+		material *DEKMaterial     // KMS-wrapped DEK; nil after rotation, protected by Sealer.mu
+		exp      time.Time        // When this key expires
+		incr     time.Duration    // How long each DEK is valid for
+		expBuf   time.Duration    // How long before expiry should we regenerate
+		now      func() time.Time // Clock source, shared with the parent Sealer
 	}
 )
 
@@ -59,8 +68,9 @@ type (
 // Returns an error if any DEK generation fails.
 func NewSealer(r *KEKRegistry, opts ...SealerOption) (*Sealer, error) {
 	sopts := &sealerOptions{
-		config: make(map[string]*KeyConfig),
-		now:    time.Now,
+		cacheSize: defaultCacheSize,
+		config:    make(map[string]*KeyConfig),
+		now:       time.Now,
 	}
 
 	for _, opt := range opts {
@@ -71,6 +81,14 @@ func NewSealer(r *KEKRegistry, opts ...SealerOption) (*Sealer, error) {
 		keys:     make(map[string]*sealerDEK),
 		registry: r,
 		now:      sopts.now,
+	}
+
+	if sopts.cacheSize > 0 {
+		c, err := lru.New[string, *DEK](sopts.cacheSize)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create DEK cache: %w", err)
+		}
+		s.dekCache = c
 	}
 
 	for k, v := range sopts.config {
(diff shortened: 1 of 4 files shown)


internal/crypto/sealer.go
```diff
@@ -112,44 +128,83 @@ func WithNowFunc(fn func() time.Time) SealerOption {
...
 	ct, err := dek.Encrypt(ctx, data)
 	if err != nil {
 		return nil, err
 	}
 
+	if material != nil {
+		return &Envelope{CipherText: ct, KeyMaterial: material}, nil
+	}
+
+	// First Seal after a rotation: call KMS outside any lock.
```
liam-lowe: nit: 

Is this technically "once per goroutine" because it's not under lock?

In a high volume ns (or account if default key is set) - is it possible we get rate limiting here.

Ie. 

1. we get a cache miss
2. under no lock we get kms details (taking `x` ms)
  3. during `x` ms period - each goroutine would also cache miss, into making an encrypt call
4. Only one goroutine would write to cache because of guards/lock

How high volume would an account/ns throughput need to be to hit KMS rate limits. It's probably unlikely to get rate limited - AWS is a pretty high threshold (I think 100k/s in major regions, unsure other cloud providers)

internal/crypto/sealer.go
```diff
@@ -112,44 +128,83 @@ func WithNowFunc(fn func() time.Time) SealerOption {
...
 	if err != nil {
 		return nil, err
 	}
 
+	// Snapshot key and material under RLock; release immediately so the
+	// encryption work happens outside the lock.
+	s.mu.RLock()
+	dek := sd.key
+	material := sd.material
+	s.mu.RUnlock()
```
liam-lowe, line 155: this pattern is potentially error prone for future users (ie forgetting to access .key and .material under lock - should sealerDEK self-lock?

liam-lowe (approved):

internal/crypto/sealer.go
pseudomuto: How would moving the lock into sealerDEK help? (I'm assuming that's what you mean by self-lock). If we made some method to return the key and material, future users could still (erroneously) access `.key` and `.material` without the lock. I'm probably missing some part of what you meant. 

To be clear neither the type, nor the fields are exported.

internal/crypto/sealer.go
pseudomuto: I checked out the rate limits for the supported providers (see below). The tricky bit is these aren't oranges to oranges. For example, the Google limit, as far as I can tell, is per project (not key), whereas Azure is per regional vault).

While these limits are high, I'm in agreement about the issue you raised for very busy namespaces. I can think of two options:

- Use a [singleflight.Group](https://pkg.go.dev/golang.org/x/sync/singleflight) to dedup and coalesce calls 
- Acquire the write lock (currently line 176) before calling Encrypt. This would ensure we only do the work once, but would put a network call in a critical region.

I'm inclined to use the singleflight group. From the docs:

> Do executes and returns the results of the given function, making sure that only one execution is in-flight for a given key at a time. If a duplicate comes in, the duplicate caller waits for the original to complete and receives the same results. The return value shared indicates whether v was given to multiple callers.

**Rate Limits per Cloud Provider**

- Amazon: [100k/s](https://docs.aws.amazon.com/kms/latest/developerguide/requests-per-second.html)
- Azure [4k/10s](https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits?source=recommendations#resource-type-vault)
- Google: [60k/s](https://docs.cloud.google.com/kms/quotas) (6M/s but 100 per op)

pseudomuto: Updated to dedup calls for the same namespace. This will allow the first call to go through and block the subsequent calls. When it succeeds (or fails) all blocked call receive the same result.

liam-lowe: > Azure [4k/10s](https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits?source=recommendations#resource-type-vault)

That azure rate limit is low (comparatively)!

> Use a [singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) to dedup and coalesce calls

I havent seen singleflight before!

> Do executes and returns the results of the given function, making sure that only one execution is in-flight for a given key at a time. If a duplicate comes in, the duplicate caller waits for the original to complete and receives the same results. The return value shared indicates whether v was given to multiple callers.

Nice - so essentially all other goroutines wait on the same result, and the result is shared. Solves the rate limiting concern!

---

Maybe this is nitpicky and a corner-case - should we dedupe on `DEK + Namespace`

I'm imagining the KMS is provider is down / slow - and refresh rate is fast.

1. DEK refresh!
1. Request comes in and hits singleflight dedupe for `ns` with `dek1`
2. Encryption hangs for 5 mins
3. for the sake of example, assume a goroutine hit the single flight wait group every second for next 5 mins
4. refresh occurs in background every 30 seconds. What would have been `encrypt(ns, dek{1,2,3,4,5,6,7,8,9,10}` - is all deduped into `dek1`
5. decrypt fails on these envelopes because dek{1..10} is always paired with key1

This is probably an absurd corner case given duration(encrypt) <<<<< refresh rate - but maybe there's a non-refresh corner case where this occurs for some minimal time. 

---

Related: How do we want to go about testing these race conditions - and ones we have caught. is performance testing enough . Have you used [synctest](https://go.dev/blog/synctest).

pseudomuto: > KMS is provider is down / slow

Down would be great (ECONNREFUSED is a gift!!). _Slow_ suuucks 😄  

**Rotation Issue:**

This is a very nice catch ✊! 

I don't think this is nitpicky at all, and I actually think you've identified a real issue with rotation that needs to be addressed. Here's my understanding, which I think lines up roughly with what you outlined (KMS doesn't even need to be slow for this to be a problem):

The problem is that `singleflight.Do(id, fn)` only executes the first goroutine's (G1) fn, all joiners discard their own closure and share the first result.

So when G2 arrives after rotation with dek = dek2 and calls `s.encryptor.Do(ns, func() { registry.Encrypt(ctx, ns, dek2) })` it joins G1's in-flight call and gets back material1. But G2's ciphertext was encrypted with dek2. The envelope it returns pairs ct(dek2) with material1 (no good).

On `Open`: decrypt material1 which yields dek1, try to decrypt ct(dek2) with dek1 and...💥💥💥

As you mentioned, I think we can resolve this by keying the on the DEK and the namespace. Since *DEK is a pointer, we can use the pointer address like `fmt.Sprintf("%s:%p", id, dek)`.

**Testing:** 

I've been adding test cases when possible and running tests with the `-race` flag...you know...good 'ol "trust me bro" 😄 

Synctest seems pretty cool, I'd be down to try that out for these cases for sure!

pseudomuto: I've updated [the key](https://github.com/temporalio/temporal-proxy/pull/14/changes#diff-7d30c19ac63aabc78433046792c3cff38f09529f373c3f7bf3c86f0e8173c2e5R175) and added a test for the case where rotation happens while an inflight KMS is in progress for the same namespace.

liam-lowe: Nice. Were you able to produce a failing test with the previous revision?

internal/crypto/sealer.go
liam-lowe: Fair - missed that! Something feels a bit off  - but I don't really have a better alternative other than not returning `sd` at all - and instead returning a point-in-time (not-shared) result so we don't need the lock at all.

internal/crypto/sealer.go
pseudomuto: Unfortunately not. Naturally I agree with the idea of proving the error before fixing it and regression tests in general. However this time I actually addressed it before writing the test. I'm confident that the issue was real and that the tests does cover it though. I'll be sure to be more methodical going forward.

liam-lowe: No problem - you're all good!

PR #15, opened 2026-05-04
[validation]: Add cert validation

pseudomuto:
Adds a new internal/validation package providing generic error types (validation.Error and validation.Errors) for representing structured validation failures. Each error carries a Subject (the thing being validated), a Field (the attribute that failed), and a Message (human-readable explanation), and the collection type supports full errors.As/errors.Is traversal via Unwrap.

Building on that I've added internal/validation/certs, which implements PEM certificate validation through a composable Validator function type and a certs.Validate entry point. Bundled validators cover the common cases: NotExpired (rejects expired and not-yet-valid certificates), IsCA, SecureAlgorithm (weak signature algorithms and cipher-suite key-type compatibility), and CAChain (chain-of-trust verification against a provided CA PEM).

### DIFF
--- a/internal/validation/certs/certs.go
+++ b/internal/validation/certs/certs.go
@@ -1,0 +1,81 @@
+package certs
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"fmt"
+
+	"github.com/temporalio/temporal-proxy/internal/validation"
+)
+
+// Validator checks a single certificate and returns a validation.Error if it fails.
+type Validator func(*x509.Certificate) error
+
+// Validate parses the PEM-encoded certificate data and runs each validator against every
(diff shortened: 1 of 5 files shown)


internal/validation/certs/certs.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/temporalio/temporal-proxy/internal/validation"
+)
+
+// Validator checks a single certificate and returns a validation.Error if it fails.
+type Validator func(*x509.Certificate) error
+
+// Validate parses the PEM-encoded certificate data and runs each validator against every
+// certificate found. It returns a validation.Errors containing all failures, or nil if all pass.
+func Validate(pemData []byte, validators ...Validator) error {
```
liam-lowe, line 17: elegant!

liam-lowe (commented):
Very elegant implementation!

PR #16, opened 2026-05-05
[KMS]: Wire up KMS providers via config

pseudomuto:
Currently the crypto package has the plumbing for DEK-based envelope encryption but no way to configure or connect real key-encryption keys (KEKs) to it.

This commit wires the full encryption path from YAML config through cloud KMS providers to the Sealer used at request time. (Interceptor still outstanding).

Support for AWS, Azure, GCP, and "local" crypto keys is included here. I also added `proxy tools gen-key` to make generating local/test keys simple.

The adapt package bridges config to crypto:

- `CryptoPolicies` opens a gocloud.dev/secrets keeper for each configured KMS URI and builds a crypto.KeyPolicy (KEK + DEK lifetime) per namespace.
- `RotateDEKs` runs a background goroutine that calls Sealer.Refresh every 30 seconds (for now) to proactively rotate DEKs before they expire.

The crypto package was refactored to unify per-namespace configuration. KeyPolicy replaces the former split between KEKRegistry (holding the KEK) and Sealer (holding a separate KeyConfig). A namespace now has exactly one KeyPolicy carrying both, making misconfiguration structurally impossible.

Encryption is optional. When omitted, no envelope encryption is applied. Similarly, the default stanza can be omitted to skip encryption when not explicitly defined for a namespace.

```yaml
encryption:
  # Fallback for namespaces not listed under overrides.
  default:
    uri: "gcpkms://projects/my-project/locations/global/keyRings/ring/cryptoKeys/key"
    duration: 20m      # how long each DEK is valid (minimum: 1m)
    renewBefore: 2m    # how far before expiry to proactively rotate

  # Per-namespace overrides; take precedence over default.
  overrides:
    my-namespace:
      uri: "awskms:///arn:aws:kms:us-east-1:123456789012:key/abc?region=us-east-1"
      duration: 1h
      renewBefore: 5m
```

Supported URI schemes:

- `gcpkms://` - Google Cloud KMS
- `awskms://` - AWS KMS
- `azurekeyvault://` - Azure Key Vault
- `base64key://` - Local / test (no external KMS required)

If necessary, this can be expanded to include some form of dynamic config, for example, namespaces that match a certain pattern use a key or patterned key URI, etc. For now, I've left this out.


dev/proxy.yaml
```diff
@@ -1,3 +1,15 @@
+encryption:
+  default:
+    # NB: Only suitable for testing/development.
+    # Run proxy tools gen-key to create
+    uri: base64key://dYz0gjYaoDrcBCAoTxg1tuYFxrbC1ksQcf1-med2H9w=
```
liam-lowe, line 5: any way we can enforce base64key is only used locally / in test

internal/adapt/kms.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// CryptoPolicies builds a [CryptoParams] from the encryption settings defined in the application config.
+func CryptoPolicies(ctx context.Context, c *config.Config, l log.Logger) (CryptoParams, error) {
+	p := CryptoParams{
+		Policies: make(map[string]crypto.KeyPolicy, len(c.Encryption.Policies)),
+	}
+
+	if dp := c.Encryption.DefaultKeyPolicy; dp.URI != "" {
+		l.Info("[adapt]: configuring default key policy", tag.String("uri", dp.URI))
```
liam-lowe: nit: could use `log.with` for `[scope]` injection - [ref](https://github.com/search?q=org%3Atemporalio+log.With+tag.Component&type=code)

might be too verbose for each function though..

internal/adapt/kms.go
```diff
@@ -0,0 +1,126 @@
...
+
+		p.Policies[ns] = pol
+	}
+
+	return p, nil
+}
+
+// RotateDEKs starts a background goroutine that calls [crypto.Sealer.Refresh] every 30 seconds
+// until ctx is cancelled, logging a warning if any rotation fails.
+func RotateDEKs(ctx context.Context, s *crypto.Sealer, l log.Logger) {
```
liam-lowe, line 79: nit: should we use `fx.Lifecycle` to manage instead of a goroutine?

dev/proxy.yaml
pseudomuto: We _could_, however this would involve something like having ENV=local|test|etc. and prevent the option of using the proxy without a cloud provider. I'm all for not promoting the use of it out of test/local, but it is kind of the operator's call IMO.

However, as is, this isn't quite right as it doesn't support rotation. Previous "versions" of the key need to be available to decrypt payloads encrypted before the rotation. Having a single static key breaks that.

One option is to, as you mentioned, prevent this from running in production-like environments. Another would be to update the static key version to support adding new keys.

internal/adapt/kms.go
pseudomuto: I'm happy to change it if you feel it's better. Is there an advantage to using lifecycle over an Invoke?

internal/adapt/kms.go
pseudomuto: I think I did this because it's easier for me to see `[adapt]:` in the message, but surely tags would be better for machines and processors. I'll update this. 👍

pseudomuto: Added `scopedLogger` func which adds pkg/component tags to the logger

dev/proxy.yaml
pseudomuto: We could also change the scheme to something like `testing://` to make it clear, and add some warning logs during startup. We could also require that an ENV var be set if any keys with that scheme exist or we bail on startup. Thoughts?

badideasforsale: > We could also change the scheme to something like testing:// to make it clear, 

I'm OK with that.

> and add some warning logs during startup.

I'm very OK with this. Potentially even requiring a dedicated startup flag. `--enable-local-testing-key` for example.

The more clear we can make it, the better, especially because this is going to be an area that researchers target and report things to us. The safer it can be by default, the better.

PR #17, opened 2026-05-27
[server]: Wire configurable listener via config package

pseudomuto:
Previously the server module took an optional `serverHostPort` named string and a Credentials value injected separately, defaulting to `:8443` and insecure if neither was supplied. There was no way to configure TLS or mTLS without writing Go code to wire it in.

This introduces an `internal/config` package that loads YAML from disk (or any `io.Reader`), expanding `${VAR}` references against the process environment. The top-level config currently carries a single `ListenConfig` (inline) holding host/port and optional TLS material. When TLS is set with a CA file the listener runs in mTLS mode; with cert+key only it serves plain TLS; without TLS it serves insecurely.

`server.Module` now depends on `*config.Config` and derives both the bind address and credentials from it, removing the `serverHostPort` named value, the separate `Credentials` injection, and the `DefaultHostPort` constant.


internal/config/listen.go
```diff
@@ -0,0 +1,22 @@
+package config
+
+type (
+	// ListenConfig defines properties for an inbound listener.
+	ListenConfig struct {
+		HostPort string     `yaml:"hostPort"`
```
liam-lowe, line 6: how are we intending on validating required args. 

Given we removed `hostPort = DefaultHostPort` - I imagine this argument is required. Does listening on a `""` port default to ephemeral?

liam-lowe (approved):
elegant!

internal/config/listen.go
pseudomuto: Haven't added validation yet 😬 Thinking aloud, I would rather not default (e.g. `:0`) and just make it mandatory.

If you're ok with it, I'll handle it in a subsequent PR

liam-lowe: agreed! 

Yep subsequent is fine!

PR #18, opened 2026-05-28
[config]: Add upstream cluster configuration

pseudomuto:
The proxy needs to know which Temporal cluster to route worker traffic to and how namespace names map between the local view that workers use and the remote names registered on the upstream. Until now there was no way to express either, so the proxy could only act as a listener with no notion of a destination.

This change adds an Upstream block to the top-level config, modeled around a single upstream cluster for now. NamespaceRules translates names in both directions: by default it wraps and unwraps a configurable Prefix/Suffix, and Overrides short-circuit that for namespaces whose names do not follow the convention (e.g. legacy names, or collapsing multiple local namespaces onto one remote).

### DIFF
--- a/dev/config.yaml
+++ b/dev/config.yaml
@@ -1,1 +1,16 @@
-hostPort: :8443
+hostPort: :8444
+tls:
+  ca: dev/certs/ca.crt
+  cert: dev/certs/server.crt
+  key: dev/certs/server.key
+
+upstream:
+  hostPort: localhost:7233
+  namespaces:
+    rules:
+      # Always add .remote on the way out, and remove it on the way in.
+      suffix: .remote
+      overrides:
+        # Makes ns2 and ns3 both use ns2.remote on the remote cluster.
+        - local: ns3
+          remote: ns2.remote
--- a/internal/config/upstream.go
+++ b/internal/config/upstream.go
@@ -1,0 +1,130 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/temporalio/temporal-proxy/pkg/validation"
+)
+
+type (
+	// Upstream describes a single upstream Temporal cluster the proxy connects
+	// workers to along with configuration for that remote cluster.
+	Upstream struct {
+		Listen     ListenConfig    `yaml:",inline"`
+		Namespaces NamespaceConfig `yaml:"namespaces"`
+	}
+
+	// NamespaceConfig groups the namespace translation rules for an upstream.
+	NamespaceConfig struct {
+		Rules NamespaceRules `yaml:"rules"`
+	}
+
+	// NamespaceRules translates namespace names between the local view that
+	// workers use and the remote names registered on the upstream cluster.
+	//
+	// The default translation is to wrap or unwrap a Prefix and Suffix:
+	// Remote("payments") returns Prefix+"payments"+Suffix, and Local of that
+	// returns "payments". When an explicit Overrides entry matches, the
+	// override takes precedence over the prefix/suffix rule.
+	NamespaceRules struct {
+		Prefix    string             `yaml:"prefix"`
+		Suffix    string             `yaml:"suffix"`
+		Overrides []NamespaceMapping `yaml:"overrides"`
+
+		localToRemote map[string]string
+		remoteToLocal map[string]string
+	}
+
+	// NamespaceMapping is one explicit local/remote namespace pair, used to
+	// short-circuit the prefix/suffix rule for namespaces whose names do not
+	// follow the convention.
+	NamespaceMapping struct {
+		Local  string `yaml:"local"`
+		Remote string `yaml:"remote"`
+	}
+)
+
+func (u *Upstream) Validate() error {
+	return validation.Validate(
+		"",
+		validation.Nested("", &u.Listen),
+		validation.Nested("namespaces", &u.Namespaces),
+	)
+}
+
+func (c *NamespaceConfig) Validate() error {
+	return validation.Validate(
+		"",
+		validation.Nested("rules", &c.Rules),
+	)
+}
+
+// Local returns the local namespace name that corresponds to remoteNS. If an
+// override matches it wins; otherwise the configured Prefix and Suffix are
+// stripped from remoteNS.
+func (r *NamespaceRules) Local(remoteNS string) string {
+	if v, ok := r.remoteToLocal[remoteNS]; ok {
+		return v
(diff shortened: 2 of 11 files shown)


internal/config/upstream.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// NamespaceRules translates namespace names between the local view that
+	// workers use and the remote names registered on the upstream cluster.
+	//
+	// The default translation is to wrap or unwrap a Prefix and Suffix:
+	// Remote("payments") returns Prefix+"payments"+Suffix, and Local of that
+	// returns "payments". When an explicit Overrides entry matches, the
+	// override takes precedence over the prefix/suffix rule.
+	NamespaceRules struct {
```
pseudomuto, line 30: This came up in our discussion with SF yesterday. Essentially, while a map from local -> remote works, it is cumbersome/error prone when you have many namespaces, and requires that new namespaces either use the remote name locally, or require a rollout of the proxy due to a config change.

I've added Prefix/Suffix fields to address that. If no explicit entry exists in Overrides, the outgoing namespace will be `Prefix + <ns> + Suffix`, and incoming would strip the prefix/suffix. This allows for a single rule to handle Temporal Cloud for example (set Suffix to `.<account-id>`).

The prefix could be useful when you have some isolation boundary like region or something. This is more common when migrating from OSS to Cloud. Multiple OSS clusters exist, all of which contain a `default` namespace. Using prefix, you could configure proxies in group A to prefix namespaces with `group-a-` so they can all be under the same account in Temporal Cloud.

dev/config.yaml
```diff
@@ -1 +1,16 @@
...
+upstream:
+  hostPort: localhost:7233
+  namespaces:
+    rules:
+      # Always add .remote on the way out, and remove it on the way in.
+      suffix: .remote
+      overrides:
+        # Makes ns2 and ns3 both use ns2.remote on the remote cluster.
+        - local: ns3
+          remote: ns2.remote
```
liam-lowe, line 16: haven't read the code yet - this seems ambiguous. 

I think there's a information loss here:

- `local: ns2` -> maps to `ns2.remote` (with `suffix: .remote`)
- `local: ns3` -> maps to `ns2.remote` (with `suffix: .remote`)

Maybe this is intentional (two source namespaces, one destination)

But I don't think it works because:

- `ns2.remote` can only map to one of the two namespaces on return (I imagine `ns3` takes precedent because of override). 

Haven't read the subsequent files in PR yet, so maybe this is premature:

How do we intend on dealing with this. Some thoughts:
1. Do we validate a lossless config on startup? Ie where overrides can be conflicting?
2. Do we initialize rejection rules on startup, because of overrides. IE reject ns2 as a local name?
3. Should rules be an ordered list, to give more control to customer. Maybe some customers want overrides before the suffix rule?
4. Should rules be a plugin, and not config driven?

internal/config/upstream.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return strings.TrimPrefix(strings.TrimSuffix(remoteNS, r.Suffix), r.Prefix)
+}
+
+// Remote returns the remote namespace name that corresponds to localNS. If an
+// override matches it wins; otherwise localNS is wrapped with the configured
+// Prefix and Suffix.
+func (r *NamespaceRules) Remote(localNS string) string {
+	if v, ok := r.localToRemote[strings.ToLower(localNS)]; ok {
```
liam-lowe: Lookup here is `strings.ToLower(localNS)` 

But it is used without casing-changes below: `fmt.Sprintf("%s%s%s", r.Prefix, localNS, r.Suffix)`

I think we should either abstract into a prereq sanitisation function- or skip `strings.ToLower(localNS)` altogether.

--

Example:

- `Payments` matches override rule for `payments`. But without an override would return `{prefix}-Payments-{suffix}`.

internal/config/upstream.go
```diff
@@ -0,0 +1,85 @@
...
+
+	var decoded raw
+	if err := unmarshal(&decoded); err != nil {
+		return err
+	}
+
+	*r = NamespaceRules(decoded)
+	r.localToRemote = make(map[string]string)
+	r.remoteToLocal = make(map[string]string)
+	for _, mapping := range r.Overrides {
```
liam-lowe, line 96: should err if the key already exists in overrides (invalid config)

dev/config.yaml
```diff
@@ -1 +1,16 @@
...
+hostPort: :8444
+tls:
+  ca: dev/certs/ca.crt
+  cert: dev/certs/server.crt
+  key: dev/certs/server.key
+
+upstream:
+  hostPort: localhost:7233
+  namespaces:
+    rules:
```
liam-lowe, line 10: As complexity of rules grows - how can we test involution?

I guess this isn't a requirement because we want to support "collapsing multiple local namespaces onto one remote"? 

But to me this seems like a failure mode. 

I think "collapsing multiple local namespaces onto one remote" should be a feature of a short-lived migration, not a long-lived worker proxy. I personally want to put more thought into how worker proxy can support temporary migration-specific configuration.

liam-lowe (approved):

dev/config.yaml
pseudomuto: As you mentioned above, I think this is an oversight. If we collapsed multiple local namespaces to a single remote, we couldn't get them back properly. I'll address this.

dev/config.yaml
pseudomuto: You're right about the info loss, I'll address that 👍 

I'm adding some validation logic now that #22 has landed. Overrides always take precedence over prefix/suffix rules.

Not sure what you mean by plugin here. What did you have in mind?

pseudomuto: Added validation to ensure unique local/remote names

internal/config/upstream.go
pseudomuto: I got rid of the ToLower calls. One option was to lower both sides, but after some digging, I found [this](https://docs.temporal.io/best-practices/managing-namespace#naming-conventions), which essentially means we should not play with the casing at all since Cloud and OSS treat casing differently.

internal/config/upstream.go
pseudomuto: This is handled via Validate (happens on start up)

dev/config.yaml
pseudomuto: Address by removing this option and validating no duplicate entries

PR #19, opened 2026-05-28
[namespace]: Add proto-walking Translator

pseudomuto:
The proxy needs to swap Temporal namespace names between the caller's local view and the upstream's remote view across every RPC

This adds pkg/namespace, whose Translator uses protoreflect to walk an arbitrary proto.Message and rewrite every string field whose proto name carries a namespace through a caller-supplied Mapper.

Three allow lists drive what is rewritten:

- `namespaceNameFields` for globally unambiguous names (namespace, workflow_namespace, parent_workflow_namespace, deleted_namespace)
- `messageScopedNamespaceFields` for fields like NamespaceInfo.name that only mean "namespace" on specific messages
- `ignoredNamespaceFields` for `*_id` siblings that share the word but are identifiers.

I added a bench test for the translation. The first one tests a NamespaceInfo object which is flat and has the field right at the root. The other uses a 100-level deep recursive message. On my laptop (pretty beefy M5) these were the results.

```bash 
$ go test -run='^$' -bench=. -benchmem -count=3 -benchtime=1s ./pkg/namespace

goos: darwin 
goarch: arm64
pkg: github.com/temporalio/temporal-proxy/pkg/namespace 
cpu: Apple M5 Max

BenchmarkTranslator_ToLocal_NamespaceInfo-18  6056732 ops/s  178.2 ns/op  56 B/op  3 allocs/op
BenchmarkTranslator_ToLocal_NamespaceInfo-18  6906356 ops/s  176.6 ns/op  56 B/op  3 allocs/op
BenchmarkTranslator_ToLocal_NamespaceInfo-18  6769731 ops/s  176.4 ns/op  56 B/op  3 allocs/op

BenchmarkTranslator_ToLocal_DeepRecursion-18  64540 ops/s    18490 ns/op  25 B/op  1 allocs/op
BenchmarkTranslator_ToLocal_DeepRecursion-18  64626 ops/s    18434 ns/op  24 B/op  1 allocs/op
BenchmarkTranslator_ToLocal_DeepRecursion-18  65556 ops/s    18450 ns/op  24 B/op  1 allocs/op

PASS
ok      github.com/temporalio/temporal-proxy/pkg/namespace      7.514s
```

> [!NOTE]
> _proto_test.go_ is a coverage guard: it walks every request and response reachable from the workflow, operator, and 
> admin service descriptors and fails with the exact field path whenever Temporal ships a new namespace-bearing string field the translator doesn't yet recognize.

### DIFF
--- a/pkg/namespace/proto_test.go
+++ b/pkg/namespace/proto_test.go
@@ -1,0 +1,99 @@
+package namespace_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	_ "go.temporal.io/api/operatorservice/v1"
+	_ "go.temporal.io/api/workflowservice/v1"
+	_ "go.temporal.io/server/api/adminservice/v1"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+
+	"github.com/temporalio/temporal-proxy/pkg/namespace"
+)
+
+// Services to walk to ensure we're not missing any namespace field coverage.
+// Don't forget the _ import!
+//
+// NB: We should consider parameterizing this by version so we could do
+// forward/backward compatibility tests.
+var servicesToTest = []protoreflect.FullName{
+	"temporal.api.workflowservice.v1.WorkflowService",
+	"temporal.api.operatorservice.v1.OperatorService",
+	"temporal.server.api.adminservice.v1.AdminService",
+}
+
+// TestNamespaceFieldCoverage walks every request and response message reachable
+// from the mentioned services and asserts that every string field whose proto
+// name carries namespace meaning is recognized by the translator. When Temporal
+// adds a new namespace field, this test fails with the exact field name to
+// investigate.
+func TestNamespaceFieldCoverage(t *testing.T) {
+	t.Parallel()
+
+	visited := map[protoreflect.FullName]bool{}
+	var missed []string
+
+	for _, svc := range servicesToTest {
+		desc, err := protoregistry.GlobalFiles.FindDescriptorByName(svc)
+		require.NoError(t, err, "service %q not registered; ensure its package is imported", svc)
+
+		sd, ok := desc.(protoreflect.ServiceDescriptor)
+		require.True(t, ok, "descriptor %q is not a service descriptor", svc)
+
+		methods := sd.Methods()
+		for i := range methods.Len() {
+			m := methods.Get(i)
+			scanMessageDescriptor(m.Input(), visited, &missed)
+			scanMessageDescriptor(m.Output(), visited, &missed)
+		}
+	}
+
+	require.Empty(t, missed,
+		"unrecognized namespace-bearing string fields found:\n  %s\n\n"+
+			"Each of these is a string field whose proto name contains \"namespace\" and is not\n"+
+			"in the translator's allowlist. Add it to namespaceNameFields in fields.go (if the\n"+
+			"field name globally denotes a namespace name), or to messageScopedNamespaceFields\n"+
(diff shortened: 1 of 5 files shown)


pkg/namespace/proto_test.go
```diff
@@ -0,0 +1,99 @@
...
+		if fd.Kind() == protoreflect.MessageKind || fd.Kind() == protoreflect.GroupKind {
+			scanMessageDescriptor(fd.Message(), visited, missed)
+			continue
+		}
+		if fd.Kind() != protoreflect.StringKind {
+			continue
+		}
+
+		field := string(fd.Name())
+		if !strings.Contains(field, "namespace") {
```
liam-lowe, line 88: > When Temporal adds a new namespace field, this test fails with the exact field name to investigate.

Only true when the namespace field contains `namespace`?

pkg/namespace/translator.go
```diff
@@ -0,0 +1,180 @@
...
+
+var (
+	// namespaceNameFields lists proto field names that, wherever they appear,
+	// carry a Temporal namespace name. Adding an entry expands the set of fields
+	// the translator rewrites.
+	namespaceNameFields = map[string]struct{}{
+		"namespace":                 {},
+		"workflow_namespace":        {},
+		"parent_workflow_namespace": {},
+		"deleted_namespace":         {},
```
liam-lowe, line 16: this assumes consistent field name meaning across all message types.

not trying to be a pain: should we map field to known message types - such that when we get field name for new message type - the tests fail and require manual review?

edit: I see `messageScopedNamespaceFields`. This is what I was intending here. Any reason not be explicit always?

pkg/namespace/translator.go
```diff
@@ -0,0 +1,180 @@
...
+		"parent_namespace_id":          {},
+		"parent_workflow_namespace_id": {},
+	}
+
+	// messageScopedNamespaceFields handles the cases where a namespace name lives
+	// in a generically named field (e.g. NamespaceInfo.name) that would be unsafe
+	// to add to the global allowlist.
+	messageScopedNamespaceFields = map[protoreflect.FullName]map[string]struct{}{
+		"temporal.api.namespace.v1.NamespaceInfo": {"name": {}},
+	}
```
liam-lowe, line 35: why shouldn't all fields be approved like this?

liam-lowe, line 35: I imagine we'll also need one of these for `ignoreMessageScopedNamespaceFields`.

pseudomuto: I think that list gets quite long, though we could generate it for sure. Would the value be that we can filter faster at runtime based solely on the message type?

liam-lowe: correctness. I think the default using field names could be a potential for error. Being explicit this way ensures we validate the meaning behind every new field, instead of just blanket assuming and approving.

pkg/namespace/translator.go
```diff
@@ -0,0 +1,180 @@
...
+		"namespace_ids":                {},
+		"parent_namespace_id":          {},
+		"parent_workflow_namespace_id": {},
+	}
+
+	// messageScopedNamespaceFields handles the cases where a namespace name lives
+	// in a generically named field (e.g. NamespaceInfo.name) that would be unsafe
+	// to add to the global allowlist.
+	messageScopedNamespaceFields = map[protoreflect.FullName]map[string]struct{}{
+		"temporal.api.namespace.v1.NamespaceInfo": {"name": {}},
```
liam-lowe, line 34: > handles the cases where a namespace name lives in a generically named field

I think we need a more robust test to catch the addition of these fields. 

I don't know exactly how. 

thinking out loud - can we: run the proxy, call every rpc endpoint, and then grep for original namespace in logs / output.

PR #20, opened 2026-05-28
[build]: Add goreleaser for binaries and images

pseudomuto:
Releases were previously a manual `go build` to a single `./proxy` binary, with no story for distributing cross-platform builds or container images.

This wires up goreleaser to handle:

- Generating amd64 and arm64 binaries for Linux and Darwin
- Building `temporalio/temporal-proxy:<latest|tag>` Docker images
- Creating a GitHub Release with changelog and build artifacts

`mise run tag` task is a small helper for cutting patch/minor/major releases (not fully functional just yet). I'll need to get Docker credentials in place to move forward.

CI runs ensure the build works, but do not publish anything. In the future pushing a signed tag will trigger a proper release.

### DIFF
--- a/.mise/tasks/tag
+++ b/.mise/tasks/tag
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+#MISE description="Create and push new tag"
+
+#USAGE arg "type" help="Release type" {
+#USAGE   choices "patch" "minor" "major"
+#USAGE }
+#USAGE flag "-s --suffix <suffix>" default="" help="Optional suffix to add to the tag"
+
+main() {
+  if ! git diff --quiet; then
+    echo "Working directory must be clean" >&2
+    exit 1
+  fi
+
+  if ! git diff --cached --quiet; then
+    echo "No staged changes allowed" >&2
+    exit 1
+  fi
+
+  local tag="$(generate_tag_name ${usage_type:?})"
+  if [ -n "${usage_suffix:-}" ]; then tag="${tag}${usage_suffix:-}"; fi
+
+  create_tag "${tag}"
+}
+
+generate_tag_name() {
+  local type="${1}"
+
+  case "${type}" in
+  patch)
+    git describe --tags --abbrev=0 2>/dev/null | awk -F. '{print $1"."$2"."$3+1}' || echo "v0.0.1"
+    ;;
+  minor)
+    git describe --tags --abbrev=0 2>/dev/null | awk -F. '{print $1"."$2+1".0"}' || echo "v0.1.0"
+    ;;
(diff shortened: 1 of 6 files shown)


.mise/tasks/tag
```diff
@@ -0,0 +1,51 @@
...
+    git describe --tags --abbrev=0 2>/dev/null | awk -F. '{print $1+1".0.0"}' | sed 's/^v/v/' || echo "v1.0.0"
+    ;;
+  esac
+}
+
+create_tag() {
+  local tag="${1}"
+  echo git tag -s "${tag}" -m "Release ${tag}"
+  echo git push origin "${tag}"
+  echo "Tag ${tag} created and pushed. GitHub Actions will handle build and release."
```
pseudomuto, line 48: echo's are cool for the moment, we don't want any tags until the release pipeline is set up

.mise/tasks/tag
```diff
@@ -0,0 +1,51 @@
...
+
+  case "${type}" in
+  patch)
+    git describe --tags --abbrev=0 2>/dev/null | awk -F. '{print $1"."$2"."$3+1}' || echo "v0.0.1"
+    ;;
+  minor)
+    git describe --tags --abbrev=0 2>/dev/null | awk -F. '{print $1"."$2+1".0"}' || echo "v0.1.0"
+    ;;
+  major)
+    git describe --tags --abbrev=0 2>/dev/null | awk -F. '{print $1+1".0.0"}' | sed 's/^v/v/' || echo "v1.0.0"
```
liam-lowe: Will `$1` will be prefixed with `v`? might mess up the `$1+1` math. 

I'm assuming `v1` will be coerced to an int. Then 1 will be added. 

Should this be:

`git describe --tags --abbrev=0 2>/dev/null | sed 's/^v//' \ | awk -F. '{print "v"$1+1".0.0"}' || echo "v1.0.0"`

pseudomuto: Yep...I tested the first two and just copy/pasta'd this... 😬

* `echo -n v1.2.3 | awk -F. '{print $1"."$2"."$3+1}'` -> `v1.2.4` ✅ 
* `echo -n v1.2.3 | awk -F. '{print $1"."$2+1"."0}'` -> `v1.3.0` ✅ 
* `echo -n v1.2.3 | awk -F. '{print $1+1"."0"."0}'` -> `1.0.0` ❌ 

Verified `echo -n v1.2.3 | sed 's/^v//' | awk -F. '{print "v"$1+1".0.0"}'` -> `v2.0.0` as desired.

pseudomuto:
Fixed the major tag logic

liam-lowe (approved):

PR #21, opened 2026-06-03
[validation]: Add rule-based validation API

pseudomuto:
Adds pkg/validation, a small package for collecting validation failures into a structured Errors value while keeping the call site explicit and type-safe.

The core types are:

* Error (a Subject/Field/Message triple)
* Errors (an errors.Is/As/Join-compatible aggregate via Unwrap)
* Rule (a function that produces Errors)
* Check[V] (a function that verifies one value of type V).

For now, the built-in checks are Required and Unique. However, I've designed the API to allow for things like `When[T any](Predicate[T], Check[T]) Rule` or `Range[T comparable](min, max T) Rule` in the future.

I also wired the new API into internal/transport/creds to handle basic cert/CA validation. We needed these checks and I needed something to test drive the validation API.

### DIFF
--- a/internal/transport/creds/insecure.go
+++ b/internal/transport/creds/insecure.go
@@ -28,3 +28,9 @@ func (c *Insecure) DialOption() (grpc.DialOption, error) {
 func (c *Insecure) ServerOption() (grpc.ServerOption, error) {
 	return grpc.Creds(insecure.NewCredentials()), nil
 }
+
+// Validate reports any configuration problems with this credential. Insecure
+// has no certificates or keys to inspect, so it always succeeds.
+func (c *Insecure) Validate() error {
+	return nil
+}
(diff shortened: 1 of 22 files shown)


internal/transport/creds/insecure.go
```diff
@@ -28,3 +28,9 @@ func (c *Insecure) DialOption() (grpc.DialOption, error) {
 func (c *Insecure) ServerOption() (grpc.ServerOption, error) {
 	return grpc.Creds(insecure.NewCredentials()), nil
 }
+
+// Validate reports any configuration problems with this credential. Insecure
+// has no certificates or keys to inspect, so it always succeeds.
+func (c *Insecure) Validate() error {
```
liam-lowe, line 34: nit: we add the `.Validate()`, but not the call site.

liam-lowe (approved):

internal/transport/creds/insecure.go
pseudomuto: image

I'm going to add it as part of config validation, but I didn't want to make this any larger

PR #22, opened 2026-06-04
[config]: Validate Config and its TLS material at startup

pseudomuto:
`Config` now exposes a `Validate` method that walks the entire config tree (simple right now) and reports problems before the proxy attempts to use it.

`ListenConfig` verifies `hostPort` is a valid TCP address, and, when supplied, TLS configuration is validated. TLS validation is offloaded to _transport/creds_, but conditionally uses `MTLS` or `TLS` based on whether or not CA is supplied.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -5,6 +5,8 @@ import (
 	"os"
 
 	"github.com/goccy/go-yaml"
+
+	"github.com/temporalio/temporal-proxy/pkg/validation"
 )
 
 type (
@@ -42,4 +44,8 @@ func LoadFile(path string) (*Config, error) {
 	defer func() { _ = f.Close() }()
 
 	return Load(f)
+}
+
+func (c *Config) Validate() error {
+	return validation.Validate("", validation.Nested("", &c.Listen))
 }
--- a/pkg/validation/checks.go
+++ b/pkg/validation/checks.go
@@ -3,10 +3,23 @@ package validation
 import (
 	"errors"
 	"fmt"
+	"net"
 )
 
 // Check verifies a single value of type V. A nil return means valid.
 type Check[V any] func(V) error
+
+// IsHostPort validates that s is a host:port pair accepted by net.ResolveTCPAddr.
+// Both "host:port" and ":port" (listen-on-all-interfaces) forms are valid.
+func IsHostPort() Check[string] {
+	return func(s string) error {
+		if _, err := net.ResolveTCPAddr("tcp", s); err != nil {
+			return errors.New("is not a valid host:port")
+		}
+
+		return nil
+	}
+}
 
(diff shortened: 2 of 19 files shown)


internal/config/config.go
```diff
@@ -43,3 +45,7 @@ func LoadFile(path string) (*Config, error) {
 
 	return Load(f)
 }
+
+func (c *Config) Validate() error {
+	return validation.Validate("", validation.Nested("", &c.Listen))
```
liam-lowe, line 50: should we be better utilizing this first argument (`""`) for configuration files:

My understanding is that this is the subject of what we're validating. We could pass a certificate name or uuid or identifier. 

I think for configuration (or maybe this applies to all nested cases) - it might be useful to print the path from root to get to the thing being identified. 

Example: some config

```yaml
name: liam
port: abcd
loadbalancer:
    port: defg
    servers:
       - name: server1
         port: aaaa
       - name: server2
         port: bbbb
```

If we validate that all `port` fields were valid ints - would we have two identical failure messages because they have the same leaf key. 

Could we concatenate the subjects (if existing subjects in nested case, then prepend)? or via some other way, pass the nested context?

pkg/validation/checks.go
```diff
@@ -3,11 +3,24 @@ package validation
...
 )
 
 // Check verifies a single value of type V. A nil return means valid.
 type Check[V any] func(V) error
 
+// IsHostPort validates that s is a host:port pair accepted by net.ResolveTCPAddr.
+// Both "host:port" and ":port" (listen-on-all-interfaces) forms are valid.
+func IsHostPort() Check[string] {
+	return func(s string) error {
+		if _, err := net.ResolveTCPAddr("tcp", s); err != nil {
```
liam-lowe, line 16: [Looks like this performs DNS resolution.](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/tcpsock.go;l=84)

I'm unsure how I feel about this. I think this is okay for a startup check "can we resolve the host provided" - but it's a little odd for a config check "is this a valid format hostport". 

I think there's potentiall two validations being conflated here: "IsHostPort" and "canResolveHostPort". Ie on any machine "example.internal.cloud.temporal.io:8080" is a valid hostport, but maybe only resolvable from my personal machine. 

At the least - I think the error message could be clearer to indicate this differentiation

In general - I think config validation shouldn't use IO and should be pure - but I'm okay for startup to validate a bit deeper.

liam-lowe (approved):

internal/config/config.go
pseudomuto: Agreed. Since right now, this only validates ListenConfig, which is inlined for YAML, blank string is the path. For example, if the TLS cert is invalid, the key will be `tls.cert` which is, as you mentioned, the path to the key.

PR #30, opened 2026-06-11
[crypto]: Add pkg/crypto for envelope encryption

pseudomuto:
Introduces pkg/crypto, which implements [envelope encryption] for protecting data at rest and in transit. Data is encrypted with a per-payload Data Encryption Key (DEK) using [AES-256-GCM], and each DEK is, in turn, wrapped with a customer-managed Key Encryption Key (KEK), typically backed by a cloud KMS. The wrapped DEK and the ID of the KEK that produced it travel together as `DEKMaterial`, so the DEK can be recovered, and the data decrypted later.

A KEKRegistry manages the available KEKs, selecting the appropriate key by namespace for encryption and by key ID for decryption. Retired keys can be registered for decryption only, so previously encrypted DEKs remain available for decryption.

[AES-256-GCM]: https://en.wikipedia.org/wiki/Galois/Counter_Mode
[envelope encryption]: https://en.wikipedia.org/wiki/Hybrid_cryptosystem#Envelope_encryption

### DIFF
--- a/pkg/crypto/dek.go
+++ b/pkg/crypto/dek.go
@@ -1,0 +1,85 @@
+package crypto
+
+import (
+	"context"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"errors"
+	"fmt"
+)
+
+const (
+	keyBytes        = 32 // 256 bits
+	materialVersion = byte(1)
+)
+
+// ErrMalformedCipherText indicates that the ciphertext was not created by a call
+// to Encrypt, or that it was otherwise tampered with.
+var ErrMalformedCipherText = errors.New("invalid ciphertext, not encrypted with a DEK")
+
+type (
+	// DEK defines a Data Encryption Key used to encrypt/decrypt payloads.
+	DEK struct {
+		key []byte
+		gcm cipher.AEAD
+	}
+
+	// DEKMaterial defines the material needed in order to decrypt a payload.
+	DEKMaterial struct {
+		Version      byte
+		KEKID        string // The ID/URI of the KEK the encrypted the DEK.
+		EncryptedDEK string // The base64-encoded encrypted DEK.
+	}
+)
+
+// NewDEK generates a new random 256-bit Data Encryption Key.
+func NewDEK() (*DEK, error) {
+	k := make([]byte, keyBytes)
+	if _, err := rand.Read(k); err != nil {
+		return nil, fmt.Errorf("failed to create DEK: %w", err)
+	}
+
+	return dekFromKey(k)
+}
+
+// Encrypt encrypts the plaintext pt using AES-256-GCM. The returned ciphertext
+// is prefixed with the randomly generated nonce.
+func (d *DEK) Encrypt(ctx context.Context, pt []byte) ([]byte, error) {
+	nonce := make([]byte, d.gcm.NonceSize())
+	if _, err := rand.Read(nonce); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/crypto/dek.go
```diff
@@ -0,0 +1,81 @@
...
+
+type (
+	// DEK defines a Data Encryption Key used to encrypt/decrypt payloads.
+	DEK struct {
+		key []byte
+		gcm cipher.AEAD
+	}
+
+	// DEKMaterial defines the material needed in order to decrypt a payload.
+	DEKMaterial struct {
```
liam-lowe, line 29: is there value here in adding a `version/schema` identifier - just for future-proofing if we need to pivot approaches (new stricter security posture, security campaign, bug fix)

pkg/crypto/kek.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// nilKEK defines a KEK that does nothing.
+	nilKEK struct{}
+)
+
+// NewKEKRegistry constructs a KEKRegistry, applying opts in order.
+// The key-ID index used by Decrypt is built after all options are applied.
+func NewKEKRegistry(opts ...KEKRegistryOption) *KEKRegistry {
+	r := &KEKRegistry{
+		defaultKey: new(nilKEK),
```
liam-lowe: is `nilKek` the correct default. 

Would we prefer to fail loudly here and require an explicit `nilKek` when required.

pkg/crypto/kek.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	for _, opt := range opts {
+		opt.apply(r)
+	}
+
+	idMap := make(map[string]KEK, len(r.keks)+len(r.retiredKeys)+1)
+	idMap[r.defaultKey.ID()] = r.defaultKey
+	for _, v := range r.keks {
+		idMap[v.ID()] = v
+	}
```
liam-lowe: we need to return an error here if ID already exists in the map - duplicate IDs would make KEKs not retrievable

pkg/crypto/kek.go
```diff
@@ -0,0 +1,180 @@
...
+
+// NewKEKRegistry constructs a KEKRegistry, applying opts in order.
+// The key-ID index used by Decrypt is built after all options are applied.
+func NewKEKRegistry(opts ...KEKRegistryOption) *KEKRegistry {
+	r := &KEKRegistry{
+		defaultKey: new(nilKEK),
+		keks:       map[string]KEK{},
+	}
+	for _, opt := range opts {
+		opt.apply(r)
```
liam-lowe: nit: should options return errors instead of panicking?

pkg/crypto/dek.go
pseudomuto: Good call. Doesn't cost much and might be useful in the future. I'll add it

pkg/crypto/kek.go
pseudomuto: The whole point of `nilKEK` was to make encryption optional (configurable). Meaning someone can use the proxy without enabling encryption, but not enable encryption and then fail-open on an error.

Because this isn't fully implemented, it's hard to see, so I've updated it to require a default kek, which simplifies it a bit.

pkg/crypto/dek.go
pseudomuto: Added `Version byte` to `DEKMaterial`

liam-lowe (approved):

PR #32, opened 2026-06-15
[rfc]: Dynamic Upstream Routing

pseudomuto:
Document the proposal for upstream routing and configuration schema to support multiple upstream clusters based on namespace.

<https://github.com/temporalio/temporal-proxy/blob/rfc-upstream-routing/rfc/routing.md>

### DIFF
--- a/rfc/routing.md
+++ b/rfc/routing.md
@@ -1,0 +1,125 @@
+# Upstream Routing
+
+How to handle routing to multiple upstream clusters based on namespace.
+
+# Background
+
+The proxy currently supports a single static upstream defined in config. However, we want a single proxy (tenant) to serve many namespaces, each potentially terminating at a different address/upstream.
+
+This can be for a number of reasons, but notably, during migrations, some teams will want to move over namespace-by-namespace rather than cluster-by-cluster. They’d point all namespaces to the original upstream (typically the OSS frontend) and switch to another one as needed based on some routing rules.
+
+> [!NOTE]
+> Cloud offers three relevant flavours of endpoint: per-namespace endpoints (`<namespace>.<account>.tmprl.cloud:7233`), regional endpoints (`<region>.aws.tmprl.cloud:7233`), and private-link endpoints (per-VPC hostnames). Operators also need to mix Cloud and self-hosted upstreams in a single proxy.
+>
+> Ideally, we’d use the namespace endpoints since they’re already set up to handle multi-region traffic. However, there are times when that isn’t possible, so we should support all cases.
+
+# Proposal
+
+We need to support multiple named upstream configs in a single proxy and select among them (per request) using a set of routing rules. These rules will have access to the incoming namespace (before and after translation) and to gRPC metadata headers.
+
+## Upstreams
+
+As part of this, we’ll need to exchange the single upstream we have today for multiple upstream definitions. Each upstream will have its `hostPort` and `tls.serverName` fields templated, allowing for the following replacements:
+
+* `{{ .LocalNamespace }}` \- The namespace before translation  
+* `{{ .RemoteNamespace }}` \- The namespace after translation  
+* `{{ .Metadata.DC }}` \- The `DC` metadata value (arbitrary)  
+* `{{ index .Metadata “x-cluster” }}` \- The `x-cluster` metadata value (arbitrary)
+
+To be clear, these are simply available, but not required. Static `hostPort/serverName` values are not a problem (see the example config [below](#configuration-schema)).
+
+To minimize runtime issues, all static `hostPort` upstreams will establish eager connections at startup. Templated values are not predetermined and therefore cannot benefit from this. Those connections will be created lazily on first use.
+
+All connections will be maintained in a shared container (TBD) since connection establishment is expensive. Each connection will be a gRPC `ClientConn,` ensuring automatic redials and shared underlying channels via connection pooling.
+
+## Routing
+
+To match incoming requests to the appropriate upstream, we’ll need to define routing rules. In proxy config, we can define a set of rules that grant access to request information, which can be used to select an upstream connection.
+
+These rules will be evaluated in the order in which they are defined. The first matching rule will be used to route the request, with one notable exception: requests without a namespace (system requests). These will be sent to a particular upstream defined as the `system` upstream.
+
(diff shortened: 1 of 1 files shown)


rfc/routing.md
```diff
@@ -0,0 +1,125 @@
+# Upstream Routing
+
+How to handle routing to multiple upstream clusters based on namespace.
+
+# Background
+
+The proxy currently supports a single static upstream defined in config. However, we want a single proxy (tenant) to serve many namespaces, each potentially terminating at a different address/upstream.
+
+This can be for a number of reasons, but notably, during migrations, some teams will want to move over namespace-by-namespace rather than cluster-by-cluster. They’d point all namespaces to the original upstream (typically the OSS frontend) and switch to another one as needed based on some routing rules.
```
pseudomuto, line 9: Practically speaking, I would likely set the proxy to route a fixed/known set of namespaces to OSS and everything else to their upstream. Then, as you work your way down the list, you end up with a single "forward all namespaces" rule, rather than a bunch of namespace-specific rules.

liam-lowe, line 9: Nice. Yes this will keep granularity in sync with how migrations via s2s proxy currently work: a cluster is connected, but namespaces are replicated per-namespace. 

This will allow us to maintain this granularity, and moves the worker pointing logic into a centralised place. 

Ie:
1. connect clusters (via s2s)
2. set namespace to passive on destination cluster (via s2s)
3. force replicate (via s2s)
4. set destination to active (via s2s)
5. point workers to destination cluster (now active) (via worker proxy)
6. remove source namespace as a replication target (via s2s)

I wonder if we can add management apis on the worker proxy (this proxy) to facilitate pointing workers from source cluster to destination cluster.  

This would enable us to fully script the e2e namespace migration process from source to destination cluster. (current gap is requiring workload owner (source owner) to update worker config and restart worker pods. Could we do this with less downtime, and via API.

liam-lowe: I think something important about proxy config - especially when migrations are involved. 

The final state configuration shouldn't have residual artefacts that a migration occurred.

rfc/routing.md
```diff
@@ -0,0 +1,125 @@
...
+
+To route incoming requests to the appropriate upstream, we’ll need to define a mapping in the proxy config. The mapping can be done using the incoming namespace and/or gRPC metadata attached to the request. A rule is defined with a matcher that specifies the namespace and the metadata values to match, and an upstream to route to.
+
+Rules use `AND` logic to match. This means that defining a rule for `namespace=test` that contains a metadata `key=value` pair requires both conditions to be true to match. While more complex logic could be added, it also introduces complexity and the risk of logic errors in the config.
+
+It is an error to define a rule without a namespace or metadata values. This will fail validation on startup.
+
+#### Namespace Rules
+
+Since namespace translation is defined per upstream and we don’t yet know which upstream we’re using, the local namespace (before translation) is used to perform the match. When a namespace is not defined on the rule, no namespace-based matching will be performed.
```
liam-lowe, line 53: How will inbound routing work for glob-based rules (where inbound is a one to many relationship)

My understanding: rules are evaluated in order at evaluation-time (not pre compiled as a bi-directional map would be).

So in the case of:

```
  rules:
    - match:
        namespace: "prod-*"
      upstream: prod

    - match:
        namespace: "*-admin"
      upstream: admin
```

A namespace `prod-admin` would map to `prod-namespace` as each rule evaluates in order.

On outbound: `prod-admin` is evaluated to match `prod-*` and `prod` is returned. 

On inbound: `prod` is evaluated to match rule `prod-*`. 

So we know the rule - how is the one-to-many translation identified? This might be later in the file, or I might have missed it.

pseudomuto: This is a good point and a potential source of confusion if not documented well. These rules are for routing to _upstreams_ and are evaluated in the order they are defined in the config. There's no namespace translation here. That would be defined per upstream (not covered in this RFC).

liam-lowe: understood! The proxy is only unidirectionally routing - so there is no inbound translation required. Thanks.

liam-lowe (approved):
not sure if you wanted approval on this, or if it was just for gathering feedback. 

If gathering feedback - i'm sure you want more than one perspective - but approving to mark my alignment.

rfc/routing.md
```diff
@@ -0,0 +1,125 @@
...
+routing:
+  default: local
+  system: cloud-regional
+
+  rules:
+    - match:
+        namespace: "prod-*"
+        headers:
+          x-tier: "gold"
+      upstream: cloud-namespace
```
robzienert, line 116: Is `namespace` required?

pseudomuto: Hmm...I had been operating under the assumption that it was, but I don't think it needs to be since the rules are evaluated in order. Just curious, or did you have a use case in mind which would be headers only? Or something else?

robzienert: Namespace naming conventions are (were?) loosely enforced, so I think we'd probably want to be able to operate off of headers.

Looking way forward: I want to be able to route Nexus operations dynamically to either TCloud or a local/on-prem cluster. We would be selecting this based on request-boundary headers (e.g. "all Nexus Operations for ServiceA in NamespaceB originating from ApplicationC are routed to on-prem cluster"). I would envision a similar behavior possible for normal namespace traffic as well.

I suspect these more fancy traffic routing capabilities may also just be something we write our own interceptors for.

pseudomuto: I can definitely see the value here. I've added some logic for this ([tests here](https://github.com/temporalio/temporal-proxy/blob/defab82cc88e7ceeb8defed99b9ec090769bf196/internal/router/mux_test.go#L30)). Essentially, you can route based on a namespace glob (simplified), request metadata, or both.

```
upstreams:
  - name: cloud
    ...
  - name: on-prem
    ...

routing:
  default: cloud
  
  rules:
    # Your example above
    - upstream: on-prem
      match:
        namespace: NamespaceB
        metadata:
          X-Service: ServiceA
          X-App: ApplicationC

    # based on namespace alone
    - upstream: on-prem
      match:
        namespace: "*-onprem"

    # or based on metadata alone
    - upstream: on-prem
      match:
        metadata:
          X-OnPrem: true
```

Of course, if there's anything more sophisticated, custom interceptors is still an option.

PR #33, opened 2026-06-16
[proxy]: Add generated WorkflowService proxy

pseudomuto:
WorkflowService implements go.temporal.io's WorkflowServiceServer by forwarding every RPC to an upstream WorkflowServiceClient. That interface has ~110 methods, each an identical pass-through, so hand-writing them is pure boilerplate that silently falls behind whenever the Temporal API adds RPCs.

Instead, cmd/codegen reflects over the WorkflowServiceServer interface and generates the implementation (service.go.tmpl) plus a reflection-driven delegation test (service_test.go.tmpl). The shared forward helper copies inbound gRPC metadata onto the outbound context so it reaches the upstream. Regenerate with 'go generate ./...' after bumping the API.

The delegation test enumerates the proxy's methods by reflection and asserts each forwards to the same-named client method and returns its response unchanged. Because an un-proxied RPC falls through to the embedded Unimplemented server, the test also serves as a freshness check: it fails by name if the generated code lags behind the interface.

For now, the generator skips non-unary signatures, so a future streaming RPC won't emit code that the unary forward helper can't express. I'll address this in a future PR.


pseudomuto:
Diff looks massive (+6k), but almost all of it is codegen and doesn't need review.

internal/proxy/service.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+//go:generate go tool -modfile=../../dev/tools.mod mockgen -destination workflow_mocks_test.go -package proxy_test -typed go.temporal.io/api/workflowservice/v1 WorkflowServiceClient
+//go:generate go run github.com/temporalio/temporal-proxy/cmd/codegen proxy -o workflow_service.go -t WorkflowService
```
pseudomuto, line 11: ^ These 2 lines are needed for every service implementation we want to generate

pseudomuto:
Well, sometimes it happens. I think this PR is already done in https://github.com/temporalio/api-go/blob/main/proxy/service.go.

api-go is automatically updated whenever the Temporal API is updated, which is a big advantage.

pseudomuto:
I've converted this to a draft while I investigate options with https://github.com/temporalio/api-go

pseudomuto:
Closing in favour of #42

PR #35, opened 2026-06-17
[connect]: Add thread-safe connection pool

pseudomuto:
Adding internal/transport/connect, with a concurrency-safe pool of gRPC client connections keyed by host.

This purpose of this will be to share grpc.ClientConn instances across the process...think _the underlying connections proxy's will use for connecting to upstreams_.

It's wired up with FX to provide a usable Pool and ensure connections are closed when shutting down the app.

### DIFF
--- a/internal/transport/connect/pool.go
+++ b/internal/transport/connect/pool.go
@@ -1,0 +1,84 @@
+package connect
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+
+	"google.golang.org/grpc"
+)
+
+var (
+	// ErrDuplicateHost is returned by Pool.Set when a connection is already
+	// registered for the given host.
+	ErrDuplicateHost = errors.New("host already defined")
+
+	// ErrHostNotFound is returned by Pool.Conn when no connection is registered
+	// for the given host.
+	ErrHostNotFound = errors.New("no connection for host")
+)
+
+type (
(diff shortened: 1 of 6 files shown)


internal/transport/connect/pool.go
```diff
@@ -0,0 +1,85 @@
...
+
+	p.conns[host] = conn
+	return nil
+}
+
+// Close shuts down every connection in the pool, joining any errors. It runs
+// at most once; subsequent calls return the same result without re-closing.
+func (p *Pool) Close() error {
+	p.closeOnce.Do(func() {
+		errs := make([]error, 0, len(p.conns))
```
liam-lowe: nit: should `p.conns` be read inside the lock

internal/transport/connect/pool_test.go
```diff
@@ -0,0 +1,131 @@
+package connect_test
```
liam-lowe, line 1: should we add a test with multiple goroutines to ensure we catch race conditions (ie defensively ensure concurrency-safe)

liam-lowe (approved):

internal/transport/connect/pool.go
pseudomuto: 🤦

internal/transport/connect/pool_test.go
pseudomuto: 👍  I added `TestPoolConcurrent`

PR #37, opened 2026-06-17
[server]: Fix potential race in Start/Stop

pseudomuto:
Server.Start runs in its own goroutine (spawned by the fx OnStart hook) where it wrote s.logger and s.cancelFunc, while Stop read both fields from the caller's goroutine during OnStop. There was no synchronization between the two, so concurrent start/stop produced a data race that `go test -race` flagged reliably in the server lifecycle tests.

Guard logger and cancelFunc with a sync.Mutex. Start derives the cancel func into a local, takes the lock only to publish the two fields and snapshot the logger, then releases it before the blocking Serve call so no lock is held across I/O. Stop snapshots both fields under the lock and does its work (cancel, GracefulStop) unlocked.

I also surfaced unexpected serve errors so if Serve fails, the proxy stops/logs appropriately.

The graceful-stop path is unchanged (Serve returns nil, nothing is logged, no shutdown is triggered), so the existing lifecycle tests still pass. Verified with `go test -race -count=5 ./internal/server/` (previously failing).

### DIFF
--- a/.mise/config.toml
+++ b/.mise/config.toml
@@ -49,7 +49,7 @@ run = ["go fix ./...", "go tool -modfile=dev/tools.mod golangci-lint fmt"]
 
 [tasks.test]
 description = "Run unit tests"
-run = "go test -cover ./..."
+run = "go test -race -cover ./..."
 
 [tasks."test:bench"]
 description = "Run bench tests"
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"net"
+	"sync"
 	"time"
 
 	"go.temporal.io/server/common/log"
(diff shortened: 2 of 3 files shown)


internal/server/server.go
```diff
@@ -98,20 +103,40 @@ func WithLogger(log log.Logger) Option {
 // the periodic health check, which runs until ctx is cancelled or [Server.Stop]
 // is called.
 func (s *Server) Start(ctx context.Context, lis net.Listener) error {
+	ctx, cancel := context.WithCancel(ctx)
```
liam-lowe, line 110: do we need to defer cancel (in general, just as best practice) but also I think there's a leak if we received an `err` in `s.grpcSvr.Serve` where s.cancelFunc isn't called. 

My understanding is that there's no `stop()` in this scenario - but imo it's good defensive programming regardless.

liam-lowe (approved):

internal/server/server.go
pseudomuto: Certainly hard to argue with as a best practice. I think there is a leak here, but it's bounded since fx would call Stop (because of the Shutdowner). However, since it's idempotent, I'll err on the side of exemplary.

pseudomuto: Added `defer cancel()`

.mise/config.toml
```diff
@@ -49,7 +49,7 @@ run = ["go fix ./...", "go tool -modfile=dev/tools.mod golangci-lint fmt"]
 
 [tasks.test]
 description = "Run unit tests"
-run = "go test -cover ./..."
+run = "go test -race -cover ./..."
```
pseudomuto, line 52: If this were in place before, we would've caught the race in dev, so I've added it. If it becomes too slow/painful, we can add `tasks."test:fast"` later, which omits the `-race` flag.

PR #39, opened 2026-06-18
[visit]: Generate visitors for Payloads, Memos, etc.

pseudomuto:
Interceptors will need to walk a request or response message and act on every Payload, Memo, or DataBlob it contains, regardless of how deeply those values are nested. Writing and maintaining that traversal by hand is tedious and drifts out of sync as the Temporal API evolves.

This adds a codegen tool that reflects over each gRPC service's request/response types, builds the message graph, and emits typed visitor functions (Payloads, Memos, DataBlobs, ...) that invoke a callback for every reachable target. These generated visitors are the building block future interceptors will use to inspect or rewrite those values.

The main motivation is that reflection happens at build time, not per request. The generated code is plain typed field access, so traversal at runtime pays no reflection cost. It also buys us safety: bumping the Temporal API and regenerating produces visitors covering the new message shapes, and anything that no longer compiles surfaces the change instead of being silently skipped.

A companion visitcheck package validates the generated visitors against a protoreflect-based oracle, confirming the generator reaches exactly the set of targets reflection would find.


pseudomuto:
Similar to #33, almost all of this diff is codegen

pseudomuto:
I've converted this to a draft while I investigate options with https://github.com/temporalio/api-go

pseudomuto:
Closing in favour of #42

PR #40, opened 2026-06-18
[server]: Warn when running with insecure credentials

pseudomuto:
The proxy silently falls back to insecure (no-TLS) credentials whenever no `tls` block is configured. That's a reasonable default for local development or when TLS is terminated at another layer, but it gives an operator no signal that inbound transport is unencrypted, which is easy to ship to production by accident.

This adds a startup warning so the insecure posture is at least visible in the logs.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -22,14 +22,17 @@ type (
 		healthSvr *health.Server
 
 		cancelFunc  context.CancelFunc
+		creds       Credentials
 		healthCheck HealthCheck
 		logger      log.Logger
 	}
 
 	// Credentials produces the [grpc.ServerOption] used to configure
-	// transport security for inbound connections.
+	// transport security for inbound connections and reports whether that
+	// transport is encrypted.
 	Credentials interface {
 		ServerOption() (grpc.ServerOption, error)
+		Encrypted() bool
 	}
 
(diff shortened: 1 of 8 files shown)


internal/server/server.go
```diff
@@ -22,14 +22,17 @@ type (
...
 		logger      log.Logger
 	}
 
 	// Credentials produces the [grpc.ServerOption] used to configure
-	// transport security for inbound connections.
+	// transport security for inbound connections and reports whether that
+	// transport is encrypted.
 	Credentials interface {
 		ServerOption() (grpc.ServerOption, error)
+		Secure() bool
```
picatz: A small-ish nit: "secure" seems to conflate "encrypted" and "authenticated"; `Encrypted() bool` might be better.

picatz (approved):

internal/server/server.go
pseudomuto: Hmm...I can see your point. 

I worry that might confuse people with the payload encryption functionality provided by the proxy. The thinking around the name was that the connection is "secure" when using TLS, and the payloads are encrypted when configured.

I'm probably overthinking this and could easily tell the difference from context.

pseudomuto: Took your suggestion and renamed it to `Encrypted` 👍

PR #41, opened 2026-06-18
[creds]: Validate certificate key sizes

pseudomuto:
TLS and mTLS certificate validation already rejected expired certs, weak signature algorithms, and key types incompatible with the configured cipher suites, but placed no lower bound on key strength. A 1024-bit RSA or 224-bit ECDSA certificate would pass validation despite being below current recommendations.

This adds a HasSufficientKeySize validator that rejects RSA keys smaller than 2048 bits and ECDSA keys smaller than 256 bits, matching NIST guidance. Key types without a size parameter (e.g. Ed25519) pass through unchecked. The validator is wired into all three certificate chains: the TLS leaf, the mTLS leaf, and the mTLS CA.

> [!NOTE]
> The 2048-bit RSA / 256-bit ECDSA minimums correspond to the 112-bit security strength baseline in NIST SP 800-57 Part 1 Rev. 5 (Table 2): https://csrc.nist.gov/pubs/sp/800/57/pt1/r5/final

### DIFF
--- a/internal/transport/creds/validation.go
+++ b/internal/transport/creds/validation.go
@@ -15,6 +15,16 @@ import (
 	"github.com/temporalio/temporal-proxy/pkg/validation"
 )
 
+const (
+	// minRSAKeyBits is the smallest RSA modulus size accepted by
+	// HasSufficientKeySize. 2048 bits is the current NIST-recommended minimum.
+	minRSAKeyBits = 2048
+
+	// minECDSAKeyBits is the smallest ECDSA curve size accepted by
+	// HasSufficientKeySize. 256 bits (e.g. P-256) is the recommended minimum.
+	minECDSAKeyBits = 256
+)
+
 var (
 	// weakAlgorithms lists signature algorithms considered cryptographically broken.
 	// SHA-1 and MD5 are vulnerable to collision attacks; MD2 is obsolete; DSA is deprecated in X.509.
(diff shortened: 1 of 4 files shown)


internal/transport/creds/validation.go
```diff
@@ -15,6 +15,16 @@ import (
 	"github.com/temporalio/temporal-proxy/pkg/validation"
 )
 
+const (
+	// minRSAKeyBits is the smallest RSA modulus size accepted by
+	// HasSufficientKeySize. 2048 bits is the current NIST-recommended minimum.
```
picatz, line 20: Non-blocking, but something to consider for a long term project: under [NIST IR 8547](https://nvlpubs.nist.gov/nistpubs/ir/2024/NIST.IR.8547.ipd.pdf), RSA-2048 keys (112 bits of security) will be deprecated after 2030.

internal/transport/creds/validation.go
```diff
@@ -214,6 +224,35 @@ func UsesSecureCertificateAlgorithm(allowedSuites ...uint16) Validator {
 	}
 }
 
+// HasSufficientKeySize returns a Validator that rejects certificates whose
+// public key is below the recommended minimum size: RSA keys must be at least
+// minRSAKeyBits and ECDSA keys at least minECDSAKeyBits. Other key types
+// (e.g. Ed25519) carry no size parameter and pass through unchecked.
+func HasSufficientKeySize() Validator {
+	return func(cert *x509.Certificate) error {
+		switch pub := cert.PublicKey.(type) {
```
picatz, line 234: Should we error on _unknown_ public key types?

picatz (approved):
Non-blocking comments to consider, otherwise LGTM 👍

internal/transport/creds/validation.go
pseudomuto: Updated to return an error unless the public key is a known type

PR #45, opened 2026-06-24
[logger]: Add zerolog-backed logger package

pseudomuto:
The proxy logged through go.temporal.io/server/common/log, which dragged in the entire Temporal server module (and its transitive deps) just for a logging interface. That is a heavy dependency to carry for the small slice of functionality actually used.

Add pkg/logger that defines a small leveled, structured Logger interface with a zerolog-backed implementation (plus noop and test loggers). The tag subpackage structures key/value pairs.

The package also exposes a configurable default logger and package-level helpers for callers that do not thread a Logger through.


pkg/logger/logger.go
```diff
@@ -0,0 +1,121 @@
...
+	case LevelDebug:
+		return zerolog.DebugLevel
+	case LevelWarn:
+		return zerolog.WarnLevel
+	case LevelError:
+		return zerolog.ErrorLevel
+	default:
+		return zerolog.InfoLevel
+	}
+}
```
liam-lowe: nit: maybe this should exist on `ZeroLogger` itself or just private  as like a `fromLevel(l level) zerolog.Level`

liam-lowe (approved):

pkg/logger/logger.go
pseudomuto: Moved to zero.go as a non-exported pure func

PR #60, opened 2026-07-02
[interceptor]: Add payload codec interceptor and wire into the proxy

pseudomuto:
The proxy forwarded WorkflowService traffic upstream with no hook to transform payloads in flight. This adds the machinery to run codec chains (compression, encryption, etc.) over payloads crossing the proxy.

The new internal/interceptor package provides `Payloads`, a gRPC unary client interceptor that encodes outbound request payloads through an ordered codec chain and decodes inbound responses in reverse, so one chain undoes its own transformations.

### DIFF
--- a/internal/interceptor/fx.go
+++ b/internal/interceptor/fx.go
@@ -1,0 +1,38 @@
+package interceptor
+
+import (
+	"fmt"
+
+	"go.uber.org/fx"
+	"google.golang.org/grpc"
+
+	"github.com/temporalio/temporal-proxy/internal/proxy"
+)
+
+// Module is the fx module that provides the proxy's outbound client
+// interceptors. It builds a single payload interceptor from the optional
+// [PayloadCodec] chain in [InterceptorParams] and provides it as a
+// []grpc.UnaryClientInterceptor for the proxy to chain onto its upstream
+// connection.
+var Module = fx.Options(fx.Provide(
+	fx.Annotate(func(p InterceptorParams) ([]grpc.UnaryClientInterceptor, error) {
+		payloads, err := Payloads(p.Codecs...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to construct payloads interceptor: %w", err)
+		}
+
+		return []grpc.UnaryClientInterceptor{payloads}, nil
+	}, proxy.UnaryInterceptorsTag),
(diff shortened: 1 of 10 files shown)


zdylag-temporal:
Is it worth explaining in the docstrings here why we want to implement our own compression/encryption as middleware rather than using default call options ala https://github.com/grpc/grpc-go/blob/master/Documentation/compression.md ?

pseudomuto:
> Is it worth explaining in the docstrings here why we want to implement our own compression/encryption as middleware rather than using default call options ala https://github.com/grpc/grpc-go/blob/master/Documentation/compression.md ?

Yeah, that makes sense. It'll likely come up again, so a quick note wouldn't hurt. I've added a docstring here. For reference here's the comment (so you don't have to search for it):

```go
// These codecs transform the Temporal payloads carried inside WorkflowService
// messages, not the gRPC frames themselves. Unlike gRPC transport compression
// (grpc.UseCompressor), which a peer decodes on arrival, these transformations
// persist: the upstream server stores the encoded bytes. That is what enables
// at-rest compression and encryption, and lets an ordered chain express
// transforms (e.g. compress-then-encrypt) that fixed transport options cannot.
```

internal/interceptor/fx.go
```diff
@@ -0,0 +1,38 @@
...
+
+		return []grpc.UnaryClientInterceptor{payloads}, nil
+	}, proxy.UnaryInterceptorsTag),
+	fx.Annotate(func(_ InterceptorParams) ([]grpc.StreamClientInterceptor, error) {
+		return nil, nil
+	}, proxy.StreamInterceptorsTag),
+))
+
+// InterceptorParams collects the fx-provided dependencies used to build the
+// proxy's interceptors. Codecs is optional; when none are supplied the payload
```
Vaughan-Temporal, line 32: In `payload.go` 

this is referenced
`// payloadCodecChain applies an ordered series of codecs to payloads.
	payloadCodecChain []PayloadCodec 
`

I know in fx if value groups are consumed you can not rely on the ordering 
[`Do not rely on the order of values inside the slice. The order is randomized.`](https://uber-go.github.io/fx/value-groups/consume.html)

However I also do not see anything returning an array of PayloadCodec. 

Mostly a call out to be aware of.

Vaughan-Temporal (commented):
Minor call outs

PR #93, opened 2026-07-25
[docs]: Document payload encryption

pseudomuto:
The proxy now supports payload encryption (envelope encryption with a KMS-wrapped DEK, plus per-Namespace overrides), but the docs never mentioned it, so a reader had no way to discover the feature or turn it on.

Add it to the top-level README feature list, and give the Cloud example an opt-in path: a commented-out encryption block in config.yaml plus a README section explaining what it does.


pseudomuto:
Screenshot 2026-07-25 at 10 03 43 AM

PR #110, opened 2026-08-01
[creds]: Exempt self-signed roots from weak-signature check

parinmaru2:
## Summary

`SecureAlgorithm` in `pkg/validation/certs` rejects any certificate whose signature algorithm is in the weak set (SHA-1, MD5, MD2, DSA). This is correct for leaf/intermediate certs but breaks when the same check is applied over a CA trust-anchor bundle: many still-valid public roots carry legacy SHA-1 self-signatures for backwards compatibility while signing SHA-256 chains today. Skip the weak-signature check for self-signed certificates (`RawIssuer == RawSubject`); leave the key-type-vs-cipher-suite check running so suite-compatibility errors still surface.

## Motivation

`internal/transport/creds/validate.go` runs `caChecks()` — `NotExpired`, `IsCA`, `SecureAlgorithm(...)`, `SufficientKeySize` — over **every certificate** in the configured CA bundle when a proxy upstream is configured in `ModeMutualTLS`. If the operator points `ca:` at the OS trust store (`/etc/pki/tls/certs/ca-bundle.crt` on RHEL 9, Mozilla NSS shipped upstream), startup fails with a cascade of `signature_algorithm: weak signature algorithm: SHA1-RSA` errors on dozens of legitimate public roots (e.g. GlobalSign Root R1, GeoTrust Global CA, DigiCert Assured ID Root CA), leaving no path to trust Temporal Cloud's public issuers without hand-curating a private CA bundle.

Observed in production: a proxy pointed at Temporal Cloud crash-looped in a Kubernetes environment because the CA bundle contained enough SHA-1-self-signed roots to trigger the rejection. Chain verification would have succeeded against any of them — the roots are trusted for their presence in the store, not for their self-signature.

## The rule

A root's self-signature is **not** consulted during chain verification. `crypto/x509.Verify` walks intermediates up to a root already in the trust pool; the root is trusted because it's *in* the pool. Its self-signature is a self-attestation, not a security claim relied on by the verifier. Rejecting a root because its self-signature is SHA-1 provides no security benefit while making a substantial fraction of every real-world CA bundle unusable.

Leaves and intermediates are unchanged: they still fail loudly on weak signatures, because those signatures **are** consulted during chain verification.

## Change

`pkg/validation/certs/certs.go`:
```go
selfSigned := bytes.Equal(cert.RawIssuer, cert.RawSubject)
if !selfSigned && weakAlgorithms[cert.SignatureAlgorithm] {
    return validation.Error{...}
}
```

The key-type-vs-cipher-suite portion of `SecureAlgorithm` runs unconditionally, so `SecureAlgorithm(rsaOnlySuite)` still rejects an ECDSA self-signed root — coverage for that is added to the tests.

Godoc on `SecureAlgorithm` explains the exemption and its reasoning.

## Test plan

`pkg/validation/certs/certs_test.go` — `TestSecureAlgorithm_WeakAlgorithm` rewritten into three subtests:

- [x] `non-self-signed cert with weak signature is rejected` — build a leaf whose Issuer differs from its Subject (distinct parent/child templates so `x509.CreateCertificate` sets `RawIssuer != RawSubject`), patch `SignatureAlgorithm = SHA1WithRSA`, assert rejection.
- [x] `self-signed root with weak signature is accepted` — self-signed root, patched `SignatureAlgorithm = SHA1WithRSA`, assert accepted.
- [x] `self-signed root still fails key-type check` — RSA self-signed root vs ECDSA-only cipher suites, assert rejection with `key type "rsa" incompatible`.

Full package suite passes:
- [x] `go vet ./...` — clean
- [x] `go test -race -count=1 ./...` — all 20 packages green, including `internal/transport/creds` which is the actual caller of `caChecks()`

## Non-goals

- Not adding any config knob to opt in or out — the exemption is unconditional because a self-signed cert being present in a `ca:` bundle *is* the operator declaring it a trust anchor.
- Not touching `NotExpired`, `IsCA`, or `SufficientKeySize` — those checks make sense for roots (expired root = ignore it; non-CA in a CA bundle = misconfig; 1024-bit root = genuinely weak).
- Not changing leaf/intermediate validation.

### DIFF
--- a/pkg/validation/certs/certs.go
+++ b/pkg/validation/certs/certs.go
@@ -189,16 +189,26 @@ func IsCA() Check {
 // it additionally rejects certificates whose public key type is incompatible
 // with every suite in that list.
 //
-// Self-signed root certificates are exempt from the signature-algorithm check:
-// a root's self-signature is not part of chain verification (roots are trusted
-// because they appear in the trust store, not because their self-signature is
-// verified), and many still-valid public roots — used to sign SHA-256 chains
-// today — carry legacy SHA-1 self-signatures. Rejecting them would make the
-// system CA bundle unusable as a trust anchor. The key-type check still runs.
+// Self-issued certificates (RawIssuer == RawSubject) are exempt from the
+// signature-algorithm check. This is a superset of self-signed: it also
+// admits CA key-rollover certs, which are self-issued but signed by a
+// different (older or newer) key than the one they certify. The exemption
+// holds regardless: a self-issued cert's own signature is never consulted
+// during chain verification — the cert is trusted (or not) based on its
+// presence in the trust store, or on its role elsewhere in the chain, not on
+// its self-attestation. Many still-valid public roots — used to sign SHA-256
+// chains today — carry legacy SHA-1 self-signatures; rejecting them would
+// make the system CA bundle unusable as a trust anchor.
+//
+// SecureAlgorithm is used both for trust-anchor validation and, via
+// leafChecks, for certificates presented in a peer's chain. The exemption
+// applies in both cases: a self-issued cert anywhere in a presented chain
+// skips the weak-signature check, for the same reason. The key-type check
+// still runs unconditionally.
 func SecureAlgorithm(allowedSuites ...uint16) Check {
 	return func(cert *x509.Certificate) error {
-		selfSigned := bytes.Equal(cert.RawIssuer, cert.RawSubject)
-		if !selfSigned && weakAlgorithms[cert.SignatureAlgorithm] {
+		selfIssued := bytes.Equal(cert.RawIssuer, cert.RawSubject)
+		if !selfIssued && weakAlgorithms[cert.SignatureAlgorithm] {
 			return validation.Error{
 				Subject: cert.Subject.CommonName,
 				Field:   "signature_algorithm",
--- a/pkg/validation/certs/certs.go
+++ b/pkg/validation/certs/certs.go
@@ -1,6 +1,7 @@
 package certs
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"crypto/rsa"
 	"crypto/tls"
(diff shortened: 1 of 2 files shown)


pkg/validation/certs/certs_test.go
```diff
@@ -250,24 +250,85 @@ func TestSecureAlgorithm(t *testing.T) {
...
+		require.NoError(t, certs.SecureAlgorithm()(cert))
+	})
+
+	t.Run("self-signed root still fails key-type check", func(t *testing.T) {
+		t.Parallel()
+
+		// The self-signed exemption applies only to the weak-signature check;
+		// the key-type check must still run. An RSA self-signed root fails a
+		// checker that only accepts ECDSA suites.
+		certPEM := testutil.RSACert(t, caTemplate())
```
pseudomuto: /nit The other two tests both assert their setup (require.Equal on RawIssuer vs RawSubject), which makes sense. If `testutil.RSACert` ever uses an issuer that doesn't match the subject, this test keeps passing while silently no longer covering the exemption, right?

pkg/validation/certs/certs.go
```diff
@@ -187,9 +188,17 @@ func IsCA() Check {
 // known-weak algorithms (SHA-1, MD5, MD2, DSA). When allowedSuites are provided
 // it additionally rejects certificates whose public key type is incompatible
 // with every suite in that list.
+//
+// Self-signed root certificates are exempt from the signature-algorithm check:
```
pseudomuto: /nit Is this accurate? For example, CA key-rollover certs have Subject = Issuer, but are signed by a different key. Perhaps we should change this wording to self-issued?

Not really a problem for this repo, but I don't think the exemption is scoped to CA bundles. `SecureAlgorithm` is also used by leafChecks ([ref](https://github.com/temporalio/temporal-proxy/blob/0198c0afef06b5f35e23af01e498ff8f283d179c/internal/transport/creds/validate.go#L66)), so a self-issued cert in the presented chain now skips the weak-signature check too.

pseudomuto (approved):
This LGTM 👍 

I ran caChecks (from transport/creds) on /etc/ssl/cert.pem (MacOS Tahoe 26.5.2) with and without this change and found more cases we should handle. 127 certs, all of which were self-issued. 40 used weak algorithms (addressed in this PR). When I ran it on this branch, I got a "x509: negative serial number" error, which seems legit.

Under non-goals, the PR states that expired roots are fine because "expired root = ignore it". But I don't believe that is correct. `ValidatePEM` accumulates errors across every cert in the file and fails the whole config, so one expired root still crash-loops the proxy.

I think this is the same underlying issue (i.e. per-cert checks over a trust-anchor bundle). First SHA-1, then expired certs, then negative serial numbers, ...).

As is, this gets us in a better place, so I'm happy to get this in and build from here.

pkg/validation/certs/certs_test.go
parinmaru2: Good catch — fixed in e8b1909. The subtest now generates the self-signed cert inline (x509.CreateCertificate with tmpl as both cert and parent) and asserts RawIssuer == RawSubject before running the check directly, so a future change to testutil.RSACert can't silently break coverage of the exemption path.

pkg/validation/certs/certs.go
parinmaru2: Good point, fixed in e8b1909. Reworded the godoc to say self-issued (RawIssuer == RawSubject) rather than self-signed root, and called out that this admits CA key-rollover certs specifically. Also renamed the local var selfSigned -> selfIssued for consistency. Added a note that SecureAlgorithm's exemption applies wherever it runs, including presented chains via leafChecks, not just trust-anchor validation — the security argument still holds since a self-issued cert's own signature is never consulted during chain verification.

PR #112, opened 2026-08-04
[examples]: Add inbound auth extension server example

pseudomuto:
The proxy can delegate the inbound authentication decision to an extension server, but nothing showed how to write one. examples/kms covers the other service an extension server can implement, so this fills the gap with a runnable counterpart under examples/auth.

The example verifies an inbound JWT against a live JWKS and then requires the token's tenant claim to name a tenant it serves. That second rule is the whole reason it exists: everything else is what the built-in jwks authenticator already does, and the README says so before showing how, so a reader who does not need a custom claim is pointed back at the built-in rather than encouraged to maintain this.

The example is deliberately incomplete in two ways, both stated in its README. The identity provider authenticates nobody, so anyone who reaches it mints a token for any tenant. The extension server is reached in plaintext with no credential, which keeps setup at zero and the subject on caller authentication; ext.WithServerAuth is left commented out with a note that enabling it also means serving TLS, since the proxy refuses to put a credential on a plaintext connection. examples/kms shows that arrangement in full.


pseudomuto:
Took a different approach with the Auth API. This example is replaced with #133

PR #119, opened 2026-08-06
Add debug logging for routing decisions

parinmaru2:
## Motivation

The proxy is opaque about routing decisions at runtime: today, only lifecycle events log. With namespace translation and per-request templated `hostPort` / TLS `serverName` resolution, an operator debugging misrouted or auth-failing traffic has no way to answer, without attaching a debugger or turning up `GRPC_GO_LOG_*`, questions like:

- Which upstream did the router pick for this method and namespace?
- What did the local namespace get translated into?
- What target host:port and TLS server name did the dynamic resolver render before dialing?

This turned up while debugging a per-account namespace-suffix issue in a downstream deployment — the pod had `LOG_LEVEL=debug` set and still surfaced zero request-path lines while every worker got `Unauthenticated` from the upstream.

## Change

Two `Debug` entries on the request path, both gated so callers that don't wire a logger emit nothing:

- `router.director.Resolve` — after a successful mux decision, logs `method`, `namespace`, chosen `upstream`, and `outcome`.
- `proxy.DynamicResolver.Resolve` — after a successful resolve, logs `upstream`, `localNamespace`, `remoteNamespace`, rendered `target`, and rendered `serverName`.

Wiring:

- `RouterParams` already had (no change to callers): an optional `Logger` that now propagates into `director` and gets a `component=router` tag.
- `ProxyParams.Logger` is now threaded through `upstreamResolver` into `DynamicResolver` via a new `WithResolverLogger(logger.Logger) ResolverOption`. Untagged callers still work.

## Sample output (at Debug)

```
DEBUG routing request  component=router method=/temporal.api.workflowservice.v1.WorkflowService/StartWorkflowExecution namespace=orders upstream=cloud outcome=match
DEBUG resolved upstream target  component=resolver upstream=cloud localNamespace=orders remoteNamespace=orders.acme target=orders.acme.example:7233 serverName=orders.acme.sni.example
```

## Tests

- `internal/proxy` — new `TestDynamicResolverLogsDebugEntry` asserts the debug entry with the exact tag set via `logger.TestLogger.ContainsEntry`; `TestDynamicResolverNoLoggerNoPanic` asserts the no-logger path is a no-op.
- `internal/router` — new `TestDirectorLogsRoutingDecision` and `TestDirectorLogsSkippedWhenLoggerUnset` assert the same for the router side.

`go vet ./...` and `go test ./...` both clean locally.

## Notes

- Only fires at `Debug`, so production defaults (`Info`) are unchanged.
- No metric changes — the existing `router.Reporter.Decision` metric is untouched.
- Log-line contents deliberately include no metadata values or headers, only routing/resolution outputs, to avoid accidentally exposing per-request user data at Debug.

### DIFF
--- a/internal/proxy/fx.go
+++ b/internal/proxy/fx.go
@@ -15,6 +15,7 @@ import (
 	"github.com/temporalio/temporal-proxy/internal/transport/connect"
 	"github.com/temporalio/temporal-proxy/pkg/crypto"
 	"github.com/temporalio/temporal-proxy/pkg/logger"
+	"github.com/temporalio/temporal-proxy/pkg/logger/tag"
 )
 
 // Module is the fx module that constructs the proxy [Server] from [ProxyParams]
@@ -75,7 +76,7 @@ var Module = fx.Options(fx.Invoke(func(p ProxyParams) error {
 			dialOpts = append(dialOpts, grpc.WithChainUnaryInterceptor(enc))
 		}
 
-		res, err := upstreamResolver(up, dialOpts)
+		res, err := upstreamResolver(up, dialOpts, p.Logger)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


parinmaru2:
@pseudomuto Added some debug log lines, these logs helped me in my local worker-proxy setup to troubleshoot connectivity issues. So thought might be useful to have these as debug logs in the upstream too.

Let me know what you think about this change? Or you see any concerns with this additional logging. Thanks.

pseudomuto (approved):
Nice! LGTM 👍

pseudomuto:
Close/open just to trigger checks

PR #143, opened 2026-08-26
[deps]: Pin go.temporal.io/api to 1.62.8

pseudomuto:
Dependabot had been bumping go.temporal.io/api on its own schedule, which drifted us ahead of what Temporal Server actually ships. Server 1.31.2 vendors api 1.62.8, so proxying against 1.63.5 meant compiling against message and service definitions no server in the field speaks yet.

Downgrade to 1.62.8 and add the module to Dependabot's ignore list so the version moves deliberately, alongside a server release, rather than whenever a new api tag lands.

This was noticed as we went to [add encryption to s2s-proxy](https://github.com/temporalio/s2s-proxy/pull/260) and had to do a little dance to pin the correct API version there.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: go.temporal.io/api
     open-pull-requests-limit: 5
     labels:
       - dependencies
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: go.temporal.io/api
```
pseudomuto, line 8: Ignoring this completely, since the goal is specific version bumps as Temporal server releases happen.

liam-lowe (approved):
thanks for this!

PR #147, opened 2026-09-02
[cloud]: Validate Cloud Namespaces

pseudomuto:
The canonical "Cloud" config derives its endpoint from the translated `namespace: {{ .RemoteNamespace }}.tmprl.cloud:7233` with a `.$TEMPORAL_ACCOUNT` suffix. Config loading expands environment variables with `os.Expand`, which substitutes an empty value for an unset one, so a blank TEMPORAL_ACCOUNT silently leaves a bare-dot suffix, and every namespace translates to "<name>.". The proxy would then dial "<name>..tmprl.cloud" and the operator saw nothing but a DNS failure. More generally, the proxy had no notion of a Cloud upstream, so a namespace Cloud would reject was indistinguishable from a valid one.

Now, when an upstream is Temporal Cloud, namespaces (post-translation) are validated. For non-templated hosts, this is done at startup to prevent mistakes. For dynamic ones, a DEBUG log is printed indicating why the namespace is invalid. Requests are still forwarded upstream for now.

> NOTE: Auto-detection means a config the proxy used to accept can now
fail to start, but only one that could never have worked against Cloud anyway.

### DIFF
--- a/internal/cloud/doc.go
+++ b/internal/cloud/doc.go
@@ -1,0 +1,8 @@
+// Package cloud holds the rules that are specific to Temporal Cloud rather than
+// to any Temporal Service.
+//
+// Today that means namespace validation. Cloud identifies a namespace as
+// "<name>.<account-id>", a shape self-hosted deployments do not impose, so
+// [ValidateNamespace] checks a string against it before the proxy uses that
+// string to address a Cloud upstream.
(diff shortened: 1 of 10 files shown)


Vaughan-Temporal (approved):
Approved. This all makes

PR #152, opened 2026-09-04
[translation]: Add a library for substituting one gRPC method for ano…

Vaughan-Temporal:
Adds Method Translation Library. Review focus should generally be on the typing and structure of this, since this is establishing a new pattern. 

The core idea here is we do full message capture instead of frame by frame capture. We can not know the details of what we are translating, so that falls upon the Actual translation integration for each message set.

### DIFF
--- a/internal/cloud/translation/interceptor.go
+++ b/internal/cloud/translation/interceptor.go
@@ -1,0 +1,122 @@
+package translation
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/temporalio/temporal-proxy/internal/rpc"
+)
+
+// Option configures the interceptor [DialOptions] installs.
+type Option func(*options)
+
+type options struct {
+	via grpc.ClientConnInterface
+}
+
+// Via sends a translated call over cc instead of continuing down the chain to
+// the connection the interceptor is installed on.
+//
+// The upstream method belongs to a different service, which the connection that
+// received the call does not serve: the caller asked a Temporal Service for
+// ListNamespaces, and only Temporal Cloud's control plane can answer it. Where
+// that service lives is as fixed as the conversions themselves, so the
+// translation carries the connection rather than the request being routed to it,
+// and a request reaching any upstream is answered the same way.
+func Via(cc grpc.ClientConnInterface) Option {
+	return func(o *options) { o.via = cc }
+}
+
+// DialOptions returns the dial options that install method translation on an
+// outbound connection. Callers fold them into the dial options for the upstream
+// connection, last, so translation is the innermost interceptor: every other
+// interceptor on the chain then sees the method and message types the caller
+// asked for rather than the substitute sent upstream.
+//
+// The result is a slice though it holds a single option today. A [Translation]
+// substitutes a unary method, so a unary interceptor is all there is to install;
+// translating a streaming method would add a stream interceptor beside it, the
+// way the namespace and Cloud-namespace helpers in internal/proxy already pair
+// the two. Keeping the slice means that arrives without changing this signature
+// or the call sites, which already spread the result.
+func DialOptions(r *Registry, opts ...Option) []grpc.DialOption {
+	return []grpc.DialOption{grpc.WithChainUnaryInterceptor(unaryClientInterceptor(r, opts...))}
+}
+
+// unaryClientInterceptor returns a unary client interceptor that replaces a call
+// to a method r translates with a call to the method it translates onto: the
+// request is converted, invoked under the upstream method, and the upstream's
+// reply is folded into the reply the caller allocated. A method r does not
+// translate is invoked unchanged, as is a call whose request or reply is not a
+// proto message. Any headers the translation declares are stamped on the
+// substituted call only.
+//
+// An upstream error is returned as it arrived, so the caller sees the upstream's
+// status rather than a translated one. A conversion that fails becomes Internal
+// via [rpc.StatusError]: the mapping is compiled in, so a failure there is a
+// proxy bug rather than anything the caller did.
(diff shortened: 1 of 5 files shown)


internal/translation/doc.go
```diff
@@ -0,0 +1,20 @@
...
+// This is distinct from the namespace translation in internal/proxy, which
+// rewrites names inside a message but leaves the method alone. The two compose:
+// install method translation as the innermost interceptor so namespace
+// translation, payload codecs, and the reflective forwarder all keep seeing the
+// method and message types the caller actually asked for.
+//
+// The package is the mechanism only; it ships no translations of its own. A
+// [Translation] names the two methods and carries the conversions between their
+// message types, so what is translated lives with the domain that needs it.
+package translation
```
pseudomuto: I wonder if this will be confused with Namespace translation? Seeing as this is really, at least for now, only for Temporal Cloud, should we put it in internal/cloud?

internal/cloud/translation/interceptor.go
```diff
@@ -0,0 +1,115 @@
...
+// that service lives is as fixed as the conversions themselves, so the
+// translation carries the connection rather than the request being routed to it,
+// and a request reaching any upstream is answered the same way.
+func Via(cc grpc.ClientConnInterface) Option {
+	return func(o *options) { o.via = cc }
+}
+
+// DialOptions returns the dial options that install method translation on an
+// outbound connection. Callers fold them into the dial options for the upstream
+// connection, last, so translation is the innermost interceptor: every other
```
pseudomuto, line 34: Elsewhere in the codebase, we note that encryption (technically, our codec stack) is the innermost interceptor, ensuring everything else can work with raw messages. I understand we're only translating a few messages here, but is the expectation that we won't add any other interceptors to this connection? If an incoming message contained a Payload anywhere, I'd reasonably expect it to be encrypted even if we're translating.

Also, as is, this will allow multiple calls to DialOptions. That intentional?

internal/translation/interceptor.go
```diff
@@ -0,0 +1,115 @@
...
+// reply is folded into the reply the caller allocated. A method r does not
+// translate is invoked unchanged, as is a call whose request or reply is not a
+// proto message. Any headers the translation declares are stamped on the
+// substituted call only.
+//
+// An upstream error is returned as it arrived, so the caller sees the upstream's
+// status rather than a translated one. A conversion that fails is Internal: the
+// mapping is compiled in, so a failure there is a proxy bug and not something
+// the caller did.
+func UnaryClientInterceptor(r *Registry, opts ...Option) grpc.UnaryClientInterceptor {
```
pseudomuto: Does this need to be exported?

internal/translation/interceptor_test.go
```diff
@@ -0,0 +1,257 @@
+package translation
```
pseudomuto: /nit the convention in this repo is to use `<package>_test` and test the public interface.

internal/translation/interceptor_test.go
```diff
@@ -0,0 +1,257 @@
...
+func TestUnaryClientInterceptorPassesUntranslatedMethodsThrough(t *testing.T) {
+	t.Parallel()
+
+	var gotMethod string
+	invoker := func(_ context.Context, method string, _, _ any, _ *grpc.ClientConn, _ ...grpc.CallOption) error {
+		gotMethod = method
+		return nil
+	}
+
+	const other = "/temporal.api.workflowservice.v1.WorkflowService/ListNamespaces"
```
pseudomuto: `const` in a function...

Image

internal/cloud/translation/interceptor.go
```diff
@@ -0,0 +1,115 @@
...
+		callOpts ...grpc.CallOption,
+	) error {
+		t, ok := r.Lookup(method)
+		if !ok {
+			return invoker(ctx, method, req, reply, cc, callOpts...)
+		}
+
+		in, inOK := req.(proto.Message)
+		out, outOK := reply.(proto.Message)
+		if !inOK || !outOK {
```
pseudomuto, line 81: Is it ever the case that req and reply are not proto.Message instances?

internal/translation/interceptor.go
```diff
@@ -0,0 +1,115 @@
...
+		if !inOK || !outOK {
+			// Nothing to convert against. The forwarder types every message from the
+			// proto registry, so this only happens on a hand-rolled call; forwarding
+			// it unchanged is closer to right than failing it.
+			return invoker(ctx, method, req, reply, cc, callOpts...)
+		}
+
+		upReq, err := t.request(in)
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
```
pseudomuto: We have rpc.StatusError for this

pseudomuto (commented):
Left a few comments here, but overall looking good!

internal/cloud/translation/interceptor.go
Vaughan-Temporal: Effectively never, atleast not in the production path. However, it is possible in the testing path side of things. 

I added `TestDialOptionsForwardsNonProtoCallsUnchanged` which forces line 81 or `if !inOk || !outOk` to evaluate to true to test this line atleast.

internal/cloud/translation/interceptor.go
Vaughan-Temporal: Good catch on this, 

Ill fix the comments in #154 

Via leaves the interceptor chain when a translation fires, so anything below translation never runs for a translated call. moving translation outermost silently stopped namespace translation applying to the reply. Put the codec below and it wouldn't run at all for these calls.

Encryption also can't block conversion: the visitor only replaces common.Payload values, never structural fields, and a conversion reads structure. For the record ListNamespaces messages carry no Payload at all, so it's a no-op today. Ran encryption + translation together end to end and it passes

pseudomuto (approved):
LGTM 👍

PR #153, opened 2026-09-04
[translation]: Answer ListNamespaces from CloudService.GetNamespaces

Vaughan-Temporal:
> **Stacked on #152.** Review that first; this diff is only the mapping.

`WorkflowService.ListNamespaces` cannot be called on Temporal Cloud on any endpoint. Cloud pins a request's namespace to
the endpoint hostname and rejects any request whose body namespace differs; `ListNamespacesRequest` has no namespace
field, so its body namespace is always empty and the check can never pass. Cloud exposes the same capability as
`CloudService.GetNamespaces` on `saas-api` — a different service, different message types, different host.

Registers that translation using the library from #152.

## The judgement calls

All the lossy decisions live in `namespaces.go` and are the thing to review:

- **State.** Cloud's eleven `ResourceState` values fold onto Temporal's three. Computed once so the deleted filter and
  the reported state cannot disagree.
- **Retention.** Cloud reports whole days; the retention ttl is left unset when it reports none, rather than claiming a
  zero retention Cloud never stated. `config` itself is always present - see below.
- **`ReplicationConfig` deliberately empty.** Cloud describes replication as regional replicas, not Temporal clusters.
  Filling it with region ids would have clients read them as cluster names. `IsGlobalNamespace` is derived from the
  replica count instead.
- **Empty, never absent.** Every sub-message a frontend would populate is allocated even when there is nothing to put in
  it. A Temporal Service builds `NamespaceInfo`, `Config` and `ReplicationConfig` on every path, so clients are written
  as though they always arrive - `temporal operator namespace list` reads `resp.ReplicationConfig.ActiveClusterName`
  with no nil check. This crashed the CLI against a live account before the fix. Guarded by
  `TestListNamespacesResponseFillsWhatAClientDereferences`.
- **`include_deleted`** has nowhere to go on the Cloud request, so it is applied to the response.
- **No namespace UUID, description or owner email** — Cloud has no equivalent, so those stay empty.

## Also

`temporal-cloud-api-version` is required (`GetNamespaces` fails `InvalidArgument` without it) and pinned to
`cloudclient.DefaultAPIVersion()`, so it moves with the module the message types come from.

`CloudService` is deliberately **not** added to `internal/services`. An earlier revision of this PR put it in the
forwardable set, on the assumption the method had to resolve in the proto registry to be callable. It does not: `Adapt`
is typed on the message types, so the upstream request and reply come from the generated package `namespaces.go`
already imports, and the substituted call is issued by a client interceptor on the outbound connection - past the
allowlist, which only ever sees the service the request named on the way in, still `WorkflowService`. Listing it would
instead have served the Cloud control plane API straight through on any proxy that did not name its own
`allowedServices`, and offered it to configuration as forwardable: a capability of its own rather than anything this
translation needs. Its name now lives beside the conversions that use it, which are the only code that knows
`CloudService` by name at all. The end-to-end test in #154 covers it - that allowlist is built from `services.Known()`,
which no longer contains it.

## Scope

+524 / −2, all of it in `internal/cloud/translation` bar the `cloud-sdk` dependency itself. **Nothing installs the registry yet, so no request is
translated.** That is #3 in the stack.

## Verified against a live Cloud account

Run end-to-end against a real Cloud account with the stack from #154 in front of it, listing namespaces through the
proxy with an unmodified `temporal` CLI:

- **The translation works.** `CloudService.GetNamespaces` answered, and every namespace came back through
  `ListNamespaces` with the right name, state and retention.
- **`v0.20.0` is correct.** The pinned `temporal-cloud-api-version` - previously called out here as the likeliest
  thing to be wrong on first contact - is what the account accepts.
- **It found a real bug**, now fixed here: the CLI segfaulted rendering the reply, because it dereferences
  `ReplicationConfig` without a nil check. See the "empty, never absent" note above.

Also confirmed on the same account, for context on what else is unreachable on a Cloud frontend: `GetClusterInfo`,
`GetSearchAttributes`, `OperatorService.ListNexusEndpoints` and `OperatorService.GetNexusEndpoint` all fail the same
way as `ListNamespaces` (`InvalidArgument: namespace mismatch between header "<ns>" and request body ""`), and the
header is derived from the endpoint hostname, so no client-side header change avoids it. `GetSystemInfo` is the one
namespace-less method Cloud allowlists. Those are follow-ups, not this PR.

---

### Stack

| | PR | |
|---|---|---|

### DIFF
--- a/internal/cloud/translation/namespaces.go
+++ b/internal/cloud/translation/namespaces.go
@@ -5,6 +5,7 @@ import (
 
 	enumspb "go.temporal.io/api/enums/v1"
 	namespacepb "go.temporal.io/api/namespace/v1"
+	replicationpb "go.temporal.io/api/replication/v1"
 	workflowservice "go.temporal.io/api/workflowservice/v1"
 	cloudservice "go.temporal.io/cloud-sdk/api/cloudservice/v1"
 	cloudnamespace "go.temporal.io/cloud-sdk/api/namespace/v1"
@@ -90,13 +91,23 @@ func listNamespacesResponse(
 // response ListNamespaces returns per namespace, with state already mapped by
 // the caller so the deleted filter and this conversion agree on it.
 //
-// Only fields Cloud actually reports are set. Cloud has no namespace UUID,
+// Only fields Cloud actually reports carry a value. Cloud has no namespace UUID,
 // description, or owner email to give, and it describes replication as regional
 // replicas rather than as the clusters ReplicationConfig names, so
 // IsGlobalNamespace is derived from how many replicas there are while
-// ReplicationConfig is left unset rather than filled with region ids a client
+// ReplicationConfig is left empty rather than filled with region ids a client
 // would read as cluster names. FailoverVersion and FailoverHistory have no
 // Cloud equivalent at all.
+//
+// Empty is not the same as absent, though, and the difference is load-bearing:
+// every sub-message a Temporal Service would populate is allocated here even
+// when there is nothing to put in it. A frontend builds NamespaceInfo, Config
+// and ReplicationConfig unconditionally on every path (the server funnels them
+// all through namespaceHandler.createResponse), so clients are written against a
+// reply where they are always present - the temporal CLI reads
+// resp.ReplicationConfig.ActiveClusterName with no nil check and dies on a nil
+// one. An empty message says "Cloud did not report this" just as well as an
+// absent one, without breaking a client that has never had to handle absence.
 func describeNamespace(
 	ns *cloudnamespace.Namespace,
 	state enumspb.NamespaceState,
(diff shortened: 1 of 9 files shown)


internal/cloud/translation/namespaces.go
```diff
@@ -0,0 +1,158 @@
...
+// listNamespacesRequest converts a ListNamespaces request into the GetNamespaces
+// request that stands in for it.
+//
+// The page token is opaque to the caller and is carried across verbatim: Cloud
+// spells it as a string where WorkflowService spells it as bytes, and the token
+// a caller sends back is one this translation handed it in the first place. The
+// namespace filter has no equivalent on the Cloud request and is applied to the
+// response instead.
+func listNamespacesRequest(req *workflowservice.ListNamespacesRequest) (*cloudservice.GetNamespacesRequest, error) {
+	return &cloudservice.GetNamespacesRequest{
```
pseudomuto, line 64: This is where we'll need to figure out how to handle filters, right?

internal/services/allowlist_test.go
```diff
@@ -113,9 +113,9 @@ func TestAllowlistServiceNames(t *testing.T) {
 		want    []string
 	}{
 		{
-			name:    "the default set names the two default services",
+			name:    "the default set names every default service",
 			allowed: services.Default(),
-			want:    []string{services.OperatorService, services.WorkflowService},
+			want:    []string{services.CloudService, services.OperatorService, services.WorkflowService},
```
pseudomuto: Is this correct? Do we want to proxy CloudService?

pseudomuto (approved):
Overall LGTM 👍

internal/cloud/translation/namespaces.go
Vaughan-Temporal: yeah, If we want to add in new filters, this is where that information would go.

PR #154, opened 2026-09-04
[dataplane]: Translate the methods Cloud does not serve on a frontend

Vaughan-Temporal:
> **Stacked on #153.** This is the first PR in the stack that changes behaviour.

Nothing installed the translation registry, so the mapping from #153 was compiled in but never reached. This installs it
on a Cloud upstream — which needs no configuration to say so. Building on #147: an upstream is already known to be
Temporal Cloud, by declaration or by its address, and translating the methods Cloud does not serve on a frontend follows
from exactly that.

## Design

**The control plane is derived from the upstream** — `cloud.APIHostPort`, the upstream's own credentials (one Temporal
Cloud API key authorises both frontend and control plane), and the dial default for TLS, which verifies against the
system roots the real control plane presents. TLS is not inherited: the control plane is a different host, so the
upstream's server name or client certificate would not apply.

Where `CloudService` lives is a fixed fact about Cloud, not a policy, so it travels with the translation rather than
being restated as routing an operator could get wrong in either direction.

**Nothing turns it on or off** — it is installed for the upstream serving Namespace-less requests, when that upstream is
Cloud, and nowhere else:

```go
if up.IsCloud() && cfg.Routing.NamespacelessUpstream() == up.Name {
```

A method Cloud refuses on a Namespace endpoint is one carrying no Namespace to pin to that endpoint, so it lands on that
upstream and nowhere else; installing on any other Cloud upstream builds a control plane connection no request can
arrive over. `Routing.NamespacelessUpstream()` mirrors `router.Mux` exactly — the system upstream when one is named, the
default otherwise — so what decides where the request goes and what decides what is installed for it cannot disagree.
The fallback is the load-bearing half: `system` is optional, and the ordinary single-upstream config names only a
`default`.

An operator who prefers Cloud's own failure to a translated answer routes Namespace-less requests at a Temporal Service
that serves them. That is the same statement, made where it belongs, and it is why there is no `enabled` flag.

**An on-prem upstream is never translated** — it is not `IsCloud()`, so the interceptor is never installed on it. The
guard is at installation rather than per-call deliberately: a runtime check on the dial target would break private-link
Cloud upstreams, whose per-VPC hostnames carry no Cloud domain and are the reason `cloud: true` exists.

## Config

Everything above needs no configuration. One optional block covers the two things detection cannot know:

```yaml
apiTranslations:
  cloudApi:               # optional
    hostPort: saas-api.tmprl.cloud:443
    credentials:
      static:
        apiKey: ${TEMPORAL_API_KEY}
```

- `hostPort:` reaches a different Cloud environment.
- `credentials:` supplies the API key an mTLS Cloud upstream has none of to inherit — the Ops API takes a key only, so
  those upstreams have nothing to pass on.

A block that nothing will consult is warned about at startup, which now covers the hybrid it used to miss: Cloud
configured, Namespace-less requests served elsewhere, override inert.

## Rebased onto the TLS default

`main` gained #165 after this stack was cut: a dialled target with no `tls` block now verifies against the system roots,
and plaintext is asked for with `insecure: true`. The second commit adapts this PR to it — the control plane's empty
`tls` block is gone (it now says nothing), `insecure` joins `tls` in what a present `cloudApi` block states
authoritatively (without it a block meaning plaintext could not say so), the rule refusing a credential on an
unencrypted hop follows #165's trigger, and the e2e fakes declare the plaintext they serve. Six tests failed against
the merge with `main` before it and none do after.

## From review

- `CloudService.IsEndpoint` is now `IsSaasAPI` — a `CloudAPI` is always the control plane, so whether it is Cloud's own
  API is the only thing left to ask.
- The `enabled` tribool is gone entirely rather than reshaped, and translation is derived instead. See **Nothing turns it
  on or off** above.
- The receiver nil checks are gone, and so are the pointers that required them. Deleting the guards on their own
  segfaults every configuration that writes no `apiTranslations:` — nil was the path those configurations took — so
  `APITranslations` and `CloudAPI` became values instead. The zero value is the block nobody wrote, and it already
  described the defaults: since #165 an absent `cloudApi` and an empty one derive an identical upstream. `Cloud()` went
  with them, having existed to spare callers a presence check there is no longer any of, and the startup warning now
  triggers on "the override says something" rather than "a block exists".

Verified against the binary, not only in tests: a dev-server upstream with a `cloudApi:` block configured answers
`ListNamespaces` itself and never calls the control plane; the same upstream marked Cloud answers from the control plane
through both `temporal operator namespace list` and `grpcurl`; and a Cloud upstream reachable only by a Namespace rule
translates nothing. `GetSystemInfo` stays with the upstream in all three.

## Two subtleties worth reviewing

**Interceptor ordering is load-bearing.** Translation is installed *last* so it is innermost. `Via` leaves the chain when
a translation fires, so anything below it would not run for a translated call. Innermost, the namespace translator above
still sees the converted reply — a configured suffix rewrites `payments.a1b2c` to `payments` with no special casing.
Move it up and that silently stops working while every other test still passes, so
`TestEndToEndNamespaceRulesApplyToATranslatedReply` guards it. Mutation-checked.

**Pool keying.** `staticResolver` keys the pool by dial target, safe "because upstream hostPorts are unique (enforced by
config)" — see its comment. Control-plane connections break that: every Cloud upstream resolves to the same address, so
two upstreams with different API keys would share whichever connection was created first, and with it the first's
credentials. `keyedResolver` keys on a derived name instead.

**A non-Cloud `cloudApi:` address warns rather than fails.** A test double or private environment legitimately carries no
Cloud domain and that is indistinguishable from a typo — matching how #147 handles a namespace it cannot validate on a
templated upstream.

The connection is not opened eagerly: translation is incidental to normal traffic, so an unreachable control plane must
not stop the proxy serving everything else.

## mTLS upstreams must configure a key

The [Cloud Ops API](https://docs.temporal.io/ops) accepts an **API key only** — unlike a namespace frontend it does not
accept mTLS. So a Cloud upstream authenticating with a client certificate has no credential to inherit, and must name an
API key under `apiTranslations.cloudApi.credentials` or its translated methods are refused. That is the case the block
exists for; documented on the type.

## Scope

+958 / −0, 9 files. `dataplane.go` is the whole integration in 96 lines. Six e2e tests drive the full stack (gateway →
router → forwarder → interceptor) against a fake CloudService.

---

### Stack

| | PR | |
|---|---|---|

### DIFF
--- a/internal/config/cloudapi.go
+++ b/internal/config/cloudapi.go
@@ -1,0 +1,147 @@
+package config
+
+import (
+	"github.com/temporalio/temporal-proxy/internal/cloud"
+	"github.com/temporalio/temporal-proxy/pkg/validation"
+)
+
+// APITranslations governs rewriting a method an upstream does not serve into the
+// one that does. It is optional and usually absent: an upstream that
+// [Upstream.IsCloud] recognizes has the methods Temporal Cloud does not serve on
+// a frontend translated automatically, over a connection derived from it.
+//
+// Enabled turns that off. It defaults to on, so an absent block and an absent
+// enabled both translate; only "enabled: false" does not. Nothing else can
+// disable it, because whether an upstream needs translation is detected rather
+// than configured - this is the operator's override of that detection, for a
+// Cloud upstream whose untranslated failure is preferred to a translated answer.
+type APITranslations struct {
+	Enabled  *bool     `yaml:"enabled"`
+	CloudAPI *CloudAPI `yaml:"cloudApi"`
+}
+
+// CloudAPI overrides how the proxy reaches Temporal Cloud's control plane, which
+// answers the methods Cloud does not serve on a namespace frontend.
+//
+// The block is optional and usually absent. An upstream that [Upstream.IsCloud]
+// recognizes gets method translation on its own, over a connection to
+// [cloud.APIHostPort] carrying that upstream's credentials - the same API key
+// authorizes both, so there is nothing more to say.
+//
+// It is required in one case. The Cloud Ops API accepts an API key only; unlike a
+// namespace frontend it does not accept mTLS. An upstream authenticating with a
+// client certificate therefore has no credential to inherit, and must name an API
+// key here or its translated methods are refused. Beyond that, configure this
+// only to reach a different Cloud environment.
+//
+// See https://docs.temporal.io/ops.
+//
+// It is deliberately not an entry in Upstreams: an upstream is a server - a
+// socket, a proxy.Server, and a routing destination - and the control plane is
+// only ever a client connection. Declaring it there would give it three things
+// it cannot use and one it should not have: routability.
+type CloudAPI struct {
+	Listen      ListenConfig      `yaml:",inline"`
+	Credentials *CredentialConfig `yaml:"credentials"`
+}
+
+// Upstream renders the control plane as an [Upstream] for the Cloud upstream
+// src, so its connection is dialled by the same resolver, TLS, and credential
+// machinery as any other rather than by a second code path. A nil receiver is
+// the unconfigured case and yields the inherited defaults, so callers need not
+// branch on whether the block is present.
+//
+// Credentials are inherited from src because a Temporal Cloud API key authorizes
+// the control plane as well as the frontend. TLS is not: the control plane is a
+// different host, so src's server name or client certificate would not apply to
+// it, and a default outbound TLS configuration is used instead.
+//
+// When this block is present its own tls is authoritative, absent included, the
+// same way it is on an upstream - an absent tls dials in plaintext. That is only
+// reachable for a control plane with no credentials, since Validate rejects
+// credentials without TLS, so a key still cannot be sent in the clear.
+//
+// The name is derived from src rather than fixed, so two Cloud upstreams with
+// different credentials get distinct connections instead of sharing whichever
+// was dialled first.
+func (c *CloudAPI) Upstream(src *Upstream) *Upstream {
+	up := &Upstream{
+		Name:        src.Name + "/cloud-api",
+		Cloud:       true,
+		Listen:      ListenConfig{HostPort: cloud.APIHostPort, TLS: &TLSConfig{}},
+		Credentials: src.Credentials,
+	}
+
+	if c == nil {
+		return up
+	}
+
+	if c.Listen.HostPort != "" {
+		up.Listen.HostPort = c.Listen.HostPort
+	}
+
+	up.Listen.TLS = c.Listen.TLS
+
+	if c.Credentials != nil {
+		up.Credentials = c.Credentials
+	}
+
+	return up
+}
+
+// IsEndpoint reports whether the configured control plane addresses Temporal
+// Cloud. It is false only when an operator pointed the block somewhere else,
+// which is legitimate for a test double or a private environment, so callers
+// report it rather than reject it.
+func (c *CloudAPI) IsEndpoint() bool {
+	if c == nil || c.Listen.HostPort == "" {
+		return true
+	}
+
+	return cloud.IsEndpoint(c.Listen.HostPort)
+}
+
+// Validate checks the control plane as it will actually be dialled, by
+// validating the [Upstream] it renders to. That covers the same ground as any
+// upstream - dial target, outbound TLS, credentials, and credentials requiring
+// TLS - without restating the rules, and checks the effective configuration
+// (the defaulted address included) rather than only the fields an operator
+// supplied.
+//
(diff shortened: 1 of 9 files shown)


internal/config/cloudapi.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return up
+}
+
+// IsEndpoint reports whether the configured control plane addresses Temporal
+// Cloud. It is false only when an operator pointed the block somewhere else,
+// which is legitimate for a test double or a private environment, so callers
+// report it rather than reject it.
+func (c *CloudAPI) IsEndpoint() bool {
```
pseudomuto: This name made sense for `cloud`, but it seems odd here. Perhaps `s/IsEndpoint/IsControlPlane/`?

internal/config/cloudapi_test.go
```diff
@@ -0,0 +1,168 @@
...
+  - name: frontend
+    hostPort: ns.acct.tmprl.cloud:7233
+    tls: {}
+    credentials:
+      static:
+        apiKey: upstream-key
+apiTranslations:
+  cloudApi:
+    hostPort: saas-api.staging.tmprl.cloud:443
+    tls: {}
```
pseudomuto: Likely no longer necessary since merging #165

internal/config/config.go
```diff
@@ -97,6 +98,10 @@ func (c *Config) Validate() error {
 		validation.Nested("metrics", &c.Metrics),
 		validation.Nested("routing", &c.Routing),
 		validation.WhenRules(func() bool { return c.Auth != nil }, validation.Nested("auth", c.Auth)),
+		validation.WhenRules(
+			func() bool { return c.APITranslations != nil },
+			validation.Nested("apiTranslations", c.APITranslations),
```
pseudomuto: Not for this PR, but I wonder if `Nested` should simply return nil if its Validator argument is nil

internal/config/cloudapi.go
```diff
@@ -0,0 +1,147 @@
...
+// [Upstream.IsCloud] recognizes has the methods Temporal Cloud does not serve on
+// a frontend translated automatically, over a connection derived from it.
+//
+// Enabled turns that off. It defaults to on, so an absent block and an absent
+// enabled both translate; only "enabled: false" does not. Nothing else can
+// disable it, because whether an upstream needs translation is detected rather
+// than configured - this is the operator's override of that detection, for a
+// Cloud upstream whose untranslated failure is preferred to a translated answer.
+type APITranslations struct {
+	Enabled  *bool     `yaml:"enabled"`
```
pseudomuto: As yes...the tribool makes an appearance! 😄

internal/config/cloudapi.go
```diff
@@ -0,0 +1,147 @@
...
+// was dialled first.
+func (c *CloudAPI) Upstream(src *Upstream) *Upstream {
+	up := &Upstream{
+		Name:        src.Name + "/cloud-api",
+		Cloud:       true,
+		Listen:      ListenConfig{HostPort: cloud.APIHostPort, TLS: &TLSConfig{}},
+		Credentials: src.Credentials,
+	}
+
+	if c == nil {
```
pseudomuto: I see a few receiver-based nil guards in this PR. I would normally caution against what I would call overly defensive code. 

In practice, in what case could `c == nil` be true, but you called `c.Upstream()` without getting an NPD panic before the method body is executed?

internal/config/cloudapi.go
```diff
@@ -0,0 +1,147 @@
...
+	return c.Upstream(&Upstream{Name: "cloudApi"}).Validate()
+}
+
+// IsEnabled reports whether method translation should be installed. An absent
+// block, or a block that does not mention enabled, translates; only an explicit
+// "enabled: false" does not. Default-on is deliberate - translation makes a
+// method work that cannot work otherwise, so an operator opts out of a fix
+// rather than into one.
+func (t *APITranslations) IsEnabled() bool {
+	return t == nil || t.Enabled == nil || *t.Enabled
```
pseudomuto: Another example here. If `t == nil`, how was `t.IsEnabled()` called?

internal/config/config.go
```diff
@@ -17,6 +17,7 @@ type (
 	// Config is the top-level proxy configuration.
 	Config struct {
 		Listen           ListenConfig        `yaml:",inline"`
+		APITranslations  *APITranslations    `yaml:"apiTranslations"`
```
pseudomuto: If this wasn't a pointer, we could likely remove all of the nil checks in this PR.

pseudomuto (approved):
A left a few comments, but overall LGTM 👍 

With #165 landed, let's just be sure that TLS applies by default like it does with all the standard upstreams (I _think_ we're good, but worth a double check).

PR #155, opened 2026-09-04
[examples]: Add a listnamespace example

Vaughan-Temporal:
> **Stacked on #154.** Docs and config only — no Go.

A runnable example for calling `WorkflowService.ListNamespaces` against Temporal Cloud through the proxy, driven by
`curl` rather than a worker so the behaviour can be checked without an SDK.

The point the README makes: **nothing in `config.yaml` asks for translation.** The upstream's `.tmprl.cloud` address is
what turns it on. That is the design from #154 working as intended, and it is worth seeing what the resulting config
does *not* contain.

Walks through:

- `ListNamespaces` succeeding via the proxy
- the same call sent straight at Cloud, failing with `namespace mismatch between header "quickstart" and request body ""`
- paging via `nextPageToken`
- `GetSystemInfo` and `DescribeNamespace` still reaching the frontend untouched
- an optional `namespaces.rules` block for short namespace names

## Scope

+248 / −0, 3 files. Verified the config loads, validates, and the proxy starts on it.

---

### Stack

| | PR | |
|---|---|---|


examples/listnamespace/config.yaml
```diff
@@ -0,0 +1,74 @@
...
+# The local gateway is plaintext with no credentials, so a client can curl it
+# directly; the proxy adds TLS and the API key on the way to Cloud.
+hostPort: 127.0.0.1:7233
+
+# Ordinary routing, unchanged. There is no rule for ListNamespaces and none is
+# needed: where CloudService lives is a fixed fact about Temporal Cloud rather
+# than a policy, so the translation carries the destination itself.
+routing:
+  default: frontend
+  system: frontend
```
pseudomuto, line 36: Not needed. When not defined, system -> default by uh...default

examples/listnamespace/config.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+upstreams:
+  # The Temporal Service for this namespace. Serves all normal traffic.
+  - name: frontend
+    hostPort: ${TEMPORAL_NAMESPACE}.${TEMPORAL_ACCOUNT}.tmprl.cloud:7233
+    tls: {}
```
pseudomuto: Can drop now that #165 landed

examples/listnamespace/config.yaml
```diff
@@ -0,0 +1,74 @@
...
+    credentials:
+      static:
+        apiKey: ${TEMPORAL_API_KEY}
+
+    # Optional: uncomment to have the namespace names in the ListNamespaces reply
```
pseudomuto, line 46: I think this is ok for now, especially for a demo. However, this highlights the need to add a Filter to the GetNamespaces call so we don't end up fetching Cloud's view of the world over the proxy's view (which could be a subset of all namespaces).

examples/listnamespace/config.yaml
```diff
@@ -0,0 +1,74 @@
...
+#                    it does without the proxy. An on-prem upstream is not Cloud,
+#                    so it is never translated and needs nothing here.
+#   cloudApi:        reach a different Cloud environment, or give the control
+#                    plane credentials or TLS the upstream does not carry.
+#
+# apiTranslations:
+#   enabled: true
+#   cloudApi:
+#     hostPort: saas-api.tmprl.cloud:443
+#     tls: {}
```
pseudomuto: /nix

examples/listnamespace/README.md
```diff
@@ -0,0 +1,181 @@
...
+`ListNamespaces`: the proxy captures it on its way to whichever upstream routing chose, rewrites it to
+`CloudService.GetNamespaces`, sends it to the Cloud API instead, and converts the reply back into a
+`ListNamespacesResponse`.
+
+Nothing in `config.yaml` asks for this. A `.tmprl.cloud` address is recognised as Temporal Cloud, and translating what
+Cloud cannot serve follows from that — the same detection that turns on Cloud Namespace validation. Where
+`CloudService` lives is a fixed fact about Temporal Cloud rather than a policy an operator picks, so the destination
+travels with the translation instead of having to be restated as configuration.
+
+The control plane is reached at `saas-api.tmprl.cloud:443` with the upstream's own API key, which authorises both. An
```
pseudomuto, line 36: We should include a link to https://docs.temporal.io/ops somewhere in here

pseudomuto (approved):

PR #158, opened 2026-09-08
[metrics]: Add fixed and metadata labels to the published series

pseudomuto:
Proxy metrics could only be sliced by dimensions the proxy itself knows:
the method, the upstream, the outcome. An operator running one proxy for
several tenants had no way to attribute traffic to them, because
(typically) the thing that identifies them travels in request metadata
the collectors never see. Nor could they attribute a series to the
installation it came from (e.g. the region or zone a proxy runs in)
without jumping through hoops.

Add metrics.labels, holding both shapes alongside the existing Temporal
Namespace control. A labels.metadata entry, written "`<header>:<name>`",
carries an inbound metadata value onto the series emitted while serving
a request: the gateway's own request metrics, the routing decision, and
the vault operations. A labels.fixed entry maps a name to a constant
stamped on every series the proxy publishes.

NB: Naming a header here is worth caring about. Its value is published
on an unauthenticated /metrics, so naming a credential header exposes
that credential, and a header the caller varies freely multiplies
cardinality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,8 @@ reaches a different upstream with no change to the Worker.
 
 - **Rule-based routing.** Route requests to different upstreams by Namespace and/or request metadata, with a system
   upstream for Namespace-less calls and a default fallback.
-- **Service allowlist.** Forward only the gRPC services you name, defaulting to `WorkflowService` and
-  `OperatorService`. Server reflection is opt-in, and a service you leave out is never forwarded.
+- **Service allowlist.** Forward only the gRPC services you name, defaulting to `WorkflowService` and `OperatorService`.
+  Server reflection is opt-in, and a service you leave out is never forwarded.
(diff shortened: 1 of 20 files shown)


pseudomuto:
@Vaughan-Temporal, I updated this based on our discussion this week. Here's what it looks like now, given the ability to define both fixed labels (like region/zone) and dynamic ones like `x-tenant-id`, etc.

```yaml
# config.yaml
...
...

metrics:
  ...
  ...
  labels:
    # Report the Temporal Namespace on the series that can name one.
    # Moved from config.metrics.namespaceLabels to here.
    namespace: true

    # Constants stamped on every series the proxy publishes.
    fixed:
      region: us-west-2
      zone: us-west-2a
      deployment: gateway-a

    # Inbound request metadata carried onto the series emitted while serving
    # a request, each written "<header>:<name>".
    metadata:
      - x-tenant-id:tenant
      - x-app-name:app
```

Vaughan-Temporal (approved):

PR #165, opened 2026-09-09
[config]: Dial upstreams over TLS by default

pseudomuto:
Turning on TLS to an upstream meant writing an empty `tls: {}` block, which reads like a no-op and is easy to leave out. The creds package already treats security as the default and takes an explicit Insecure option for plaintext which seems like a good practice.

Move the decision to ListenConfig:

* A _**dialed**_ target with no TLS block now verifies its peer against the system root pool, so plaintext has to be asked for.

* A _**listener**_ has nothing to present and keeps serving plaintext until a TLS block defines a certificate. In effect, insecure is default for the gateway, because there is no known default TLS cert.

Credentials are still refused on a plaintext hop (gRPC freebie), but the trigger moves from "no tls block" to "insecure", which makes a credential over system-root TLS valid.

> [!IMPORTANT]
>
> This breaks existing configs: a plaintext upstream or extension server
> must now say `insecure: true`. Omitting it fails the proxy's start
> instead of downgrading silently.

### DIFF
--- a/examples/cloud/config.yaml
+++ b/examples/cloud/config.yaml
@@ -27,7 +27,6 @@ upstreams:
   # so the same config serves any number of namespaces.
   - name: cloud
     hostPort: "{{ .RemoteNamespace }}.tmprl.cloud:7233"
-    tls: {}
     namespaces:
       rules:
         # Rewrite the short local name to the Cloud fully-qualified name,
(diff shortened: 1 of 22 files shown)


examples/cloud/config.yaml
```diff
@@ -27,7 +27,6 @@ upstreams:
   # so the same config serves any number of namespaces.
   - name: cloud
     hostPort: "{{ .RemoteNamespace }}.tmprl.cloud:7233"
-    tls: {}
```
pseudomuto, line 30: ^ this is exactly what I'm trying to remove

Vaughan-Temporal (approved):

PR #166, opened 2026-09-10
[config]: Default cacheSize instead of disabling the DEK cache

Vaughan-Temporal:
`crypto.NewVault` defaults the DEK cache to 100, but `createVault` passes
`crypto.WithCacheSize(c.Encryption.CacheSize)` unconditionally (`internal/kms/fx.go:155`) and `config.Load` never
defaulted that field. An absent `cacheSize` therefore arrived as zero — which `WithCacheSize` documents as "disable" —
and validation accepted it, because the rule is `GTE(0)`.

So the config layer silently overrode the vault's own default, and every payload opened cost a KEK unwrap: a KMS call,
or a call to the operator's extension server, on the latency-sensitive path Workers poll. `examples/kms/config.yaml`
omits the field, so the example an operator copies shipped disabled.

## The change

- **`CacheSize` becomes a `*int`.** Zero is meaningful here, and a plain `int` cannot separate "wrote nothing" from
  "wrote zero". Absent now inherits `crypto.DefaultCacheSize`; zero still disables the cache, but has to be written
  down. (`cmp.Or`, used for the metrics defaults, cannot make that distinction — as its own test notes.)
- **`crypto.DefaultCacheSize` is exported**, so the config layer inherits the vault's number rather than restating it.
- **Validation reads the effective value** via `Encryption.DEKCacheSize()`: a negative size still fails, an absent one
  passes, and no pointer is dereferenced.
- **A disabled cache is logged at startup**, because the metrics cannot report it. `dek_cache_hits_total` and
  `dek_cache_misses_total` are created eagerly, so both sit at zero whether the cache is off or merely idle, and an
  operator debugging KMS latency reads "0 hits, 0 misses" as "unused" rather than "absent".
  
### Behavior
| Config              | Effective size                 | Behaviour                                       |
| ------------------- | ------------------------------ | ----------------------------------------------- |
| `cacheSize` absent  | `crypto.DefaultCacheSize` (100) | cache on                                        |
| `cacheSize:` (null) | 100                            | cache on — YAML null is a nil pointer, as absent |
| `cacheSize: 0`      | 0                              | cache off, with a startup warning                |
| `cacheSize: 50`     | 50                             | cache on, 50 entries                             |
| `cacheSize: -1`     | —                              | rejected by validation (`GTE(0)`)                |

## Tests

- `TestModule_OmittedCacheSizeStillCachesDEKs` — absent `cacheSize`, seal once, open twice, assert the second open hit
  the cache. Mutation-checked: forcing the size back to 0 fails it.
- `TestModule_ZeroCacheSizeDisablesTheCache` — asserts the counter's value rather than the family's absence, since eager
  creation means the family is always there.
- `TestEncryption_DEKCacheSize`, `TestLoad_DEKCacheSizeDefault` — absent, zero, set and nil-receiver, at both the struct
  and YAML levels.

### DIFF
--- a/internal/config/encryption.go
+++ b/internal/config/encryption.go
@@ -22,13 +22,14 @@ var validKeySchemes = append(crypto.DefaultSchemes(), extensionKeyScheme)
 type (
 	// Encryption configures envelope encryption of payloads. When Enabled, a
 	// Default key policy is required and governs how DEKs are provisioned and
-	// rotated. CacheSize bounds the in-memory DEK cache. Overrides maps a
-	// namespace to a key policy that supersedes Default for that namespace; the
-	// keys are pre-translation (local) namespace names, matching the namespace
-	// the vault seals under at request time.
+	// rotated. CacheSize bounds the in-memory DEK cache; see [Encryption.DEKCacheSize]
+	// for what an absent one means. Overrides maps a namespace to a key policy
+	// that supersedes Default for that namespace; the keys are pre-translation
+	// (local) namespace names, matching the namespace the vault seals under at
+	// request time.
 	Encryption struct {
 		Enabled   bool                 `yaml:"enabled"`
-		CacheSize int                  `yaml:"cacheSize"`
+		CacheSize *int                 `yaml:"cacheSize"`
 		Default   *KeyPolicy           `yaml:"default"`
 		Overrides map[string]KeyPolicy `yaml:"overrides"`
 	}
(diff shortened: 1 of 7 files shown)


internal/config/encryption.go
```diff
@@ -46,12 +47,33 @@ type (
...
+// cache, so every Open unwraps its DEK through the KEK.
+//
+// The distinction is why the field is a pointer. Zero is a meaningful value here
+// and a plain int cannot tell an operator who wrote nothing from one who wrote
+// zero - so an absent field would read as "disable the cache" and silently
+// override the vault's own default, turning every payload the proxy opens into a
+// KMS round trip. Absent means "no opinion", and disabling the cache has to be
+// written down.
+func (e *Encryption) DEKCacheSize() int {
+	if e == nil || e.CacheSize == nil {
```
pseudomuto, line 61: I think we can drop the `e == nil` check.

pseudomuto (approved):
Nice catch! 🐑 🇮🇹

internal/config/encryption.go
Vaughan-Temporal: AI is going crazy on this new insight, it technically can catch a bug but structurally is not needed. Im sorry for making you comment on all of these

PR #168, opened 2026-09-14
[metrics]: Add a Namespace label control

pseudomuto:
The namespace label on the encryption metrics is unbounded. Its value
comes from the request, so every distinct Namespace a Client names
becomes another series, and the config gave an operator no way to say
no.

Add metrics.namespaceLabels, which decides whether metrics that can
name a Temporal Namespace report it (off by default). The label is
always declared either way, and only its value is dropped, so a query
that already ignores it reads the same before and after turning it on.

NB: Namespace stays the pre-translation local name, which is the key an
operator already writes under encryption.overrides.

Breaking: Rename the three *_duration_secs histograms to
*_duration_seconds to follow the Prometheus convention.


Vaughan-Temporal (approved):

pseudomuto:
@Vaughan-Temporal, as discussed offline, I've dropped the deployment tag. Otherwise, this is the same as what you review/approved.

PR #170, opened 2026-09-15
[server]: Report liveness from a loopback Health/Watch call

Vaughan-Temporal:
## Problem

The gateway refreshed its serving status from a stub that always returned
SERVING, so the health service only proved the process could accept a
connection.

The gateway installs three stream interceptors and no unary ones, so the unary
`Check` a probe runs never traverses the chain every forwarded request goes
through. A wedge in that chain — a deadlocked interceptor, an extension server
that stopped answering — left the gateway reporting SERVING while every request
hung.

## What changed

The stub is replaced by a check that calls `grpc.health.v1.Health/Watch` and
reads one message. `Watch` because it is a locally registered *streaming*
method, so calling it runs the stream interceptor chain.

Any answer counts, including a rejection — the check never needs to
authenticate, only to be answered. Silence is the failure.

The call never leaves the process. A second, credential-free `grpc.Server` is
assembled from the same options as the real one, carrying the same interceptor
chain and the same health service, and serves an in-process listener only the
check holds. So there is no address to learn, no handshake and no credential,
and the check behaves the same under plaintext, TLS and mutual TLS. It is a
second server rather than a second listener because transport credentials
belong to a `grpc.Server`.

What that gives up is accepting on the real listener and the handshake above
it, which belong to the probe's own `tcpSocket` or `grpc` handler.

Opt-in through `server.WithLoopbackHealthCheck`; per-upstream proxies keep the
stub. No hysteresis — a probe's `failureThreshold` already debounces.

## Configuration

```yaml
health:
  interval: 30s # how often the check runs
  timeout: 5s   # bounds one run
```

Both optional and defaulted as shown, with the timeout validated to be shorter
than the interval. There is no off switch: the check only matters to a probe
pointed at the health service.

## Testing

- A wedged stream interceptor chain flips the status to NOT_SERVING and back,
  end to end through `dataplanetest`.
- A server whose listener demands client certificates reports SERVING from its
  own check.
- An interceptor that rejects every stream still leaves the status SERVING, and
  records that it ran.
- A blocked chain is bounded by the timeout, not the interval.
- The `Watch` stream is cancelled after its first message, so the watcher
  `health.Server` registers per stream does not leak every interval.
- Against `dev/config.yaml`, which is mutual TLS: at `health.timeout: 1ms` the
  gateway logs `The server did not answer its own liveness check`; at `1s` it
  logs nothing and `Health/Check` answers SERVING over mTLS.


internal/config/health.go
```diff
@@ -0,0 +1,90 @@
...
+	Interval time.Duration `yaml:"interval"`
+	Timeout  time.Duration `yaml:"timeout"`
+}
+
+// CheckEnabled reports whether the liveness check runs: the configured value
+// when there is one, and true when the field is absent. Turning the check off
+// leaves the health service in place, reporting SERVING on the same cadence, so
+// a probe configured against it keeps working.
+func (h *Health) CheckEnabled() bool {
+	if h == nil || h.Enabled == nil {
```
pseudomuto: I believe we've had this conversation 😄 `Config.Health` isn't a pointer, so it will never be `nil` in our case, right?

internal/config/health.go
```diff
@@ -0,0 +1,90 @@
...
+// Enabled turns the check on or off, Interval is how often it runs, and Timeout
+// bounds one run. Every field is optional: see [Health.CheckEnabled],
+// [Health.CheckInterval] and [Health.CheckTimeout] for what an absent one means.
+//
+// Enabled is a pointer because false is a meaningful value and the default is
+// true, so a plain bool could not tell an operator who wrote nothing from one
+// who wrote false, and an absent field would read as a request to turn the check
+// off.
+type Health struct {
+	Enabled  *bool         `yaml:"enabled"`
```
pseudomuto: Do we ever want to disable a health check? Operators can choose whether or not to call it in their environment.

internal/config/health.go
```diff
@@ -0,0 +1,90 @@
...
+
+// CheckInterval is how often the check runs: the configured interval when there
+// is one, and [defaultHealthInterval] when the field is absent. A zero interval
+// would spin the refresh loop, so this is the only reader of the field.
+func (h *Health) CheckInterval() time.Duration {
+	if h == nil {
+		return defaultHealthInterval
+	}
+
+	return cmp.Or(h.Interval, defaultHealthInterval)
```
pseudomuto, line 34: Should we avoid negatives? Something like:

`cmp.Or(max(time.Duration(0), h.Interval), defaultHealthInterval)`

If yes, we should probably do it in CheckTimeout as well

EDIT: I see now it's done in Validate()...not really necessary in that case.

internal/dataplane/dataplane.go
```diff
@@ -184,6 +186,9 @@ func New(ctx context.Context, cfg *config.Config, opts ...Option) (*Dataplane, e
 		// Health entries come from the allowlist, so what the gateway reports a
 		// status for is exactly what it will forward.
 		server.WithHealthServices(o.allowlist.ServiceNames()...),
+		// Built before anything is bound, since the listener is created during
+		// Start; Start hands it the address then.
+		server.WithHealthCheck(dp.health),
```
pseudomuto: In server.runHealthCheck, we iterate over the health services we support (previous code line) and set the status (for the respective server) accordingly. To make sure I understand this correctly, can you explain what this health check does compared with the others?

internal/dataplane/loopback.go
```diff
@@ -0,0 +1,229 @@
...
+		mu   sync.Mutex
+		addr net.Addr
+	}
+)
+
+// newLoopbackCheck builds the check for cfg's gateway. It is constructed before
+// anything is bound, because the gateway is built during New and the listener is
+// created during Start, so the address arrives later via setAddr.
+//
+// The check turns itself off for a gateway that requires client certificates.
```
pseudomuto: Is this correct? If we use mTLS, what do we use for K8S health (liveness/readiness) checks?

internal/dataplane/dataplane.go
Vaughan-Temporal: Previously the HealthCheck returned serving no matter what. This loopback check is more of a liveness check than a health check, A request has to flow through the interceptors to ensure work can be done. 

The line above, health checks the upstreams, which is slightly different than the responisibility of the liveness check IMO.

internal/config/health.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Health configures the liveness check behind the gateway's serving status.
+// Interval is how often it runs and Timeout bounds one run; both are optional,
+// and see [Health.CheckInterval] and [Health.CheckTimeout] for what an absent
+// one means.
+//
+// There is no switch to turn the check off. The health service is registered
+// either way, and whether the check's answer has any consequence is already an
+// operator's decision: it only matters to a probe pointed at the service, so not
+// configuring that probe is the off switch.
```
pseudomuto: /nit likely don't need this paragraph

internal/dataplane/dataplane.go
```diff
@@ -184,6 +184,14 @@ func New(ctx context.Context, cfg *config.Config, opts ...Option) (*Dataplane, e
 		// Health entries come from the allowlist, so what the gateway reports a
 		// status for is exactly what it will forward.
 		server.WithHealthServices(o.allowlist.ServiceNames()...),
+		// The entries above are what a status is reported under; this is what
+		// decides the status they all carry. Every entry reports one value, and
+		// until now that value came from a stub that returned SERVING
+		// unconditionally, making the health service a heartbeat for grpc-go's
+		// transport. This replaces the stub with a check that calls the gateway's
+		// own Health/Watch, so the value reflects whether a request can still
+		// travel the stream interceptor chain.
```
pseudomuto: After the first sentence, this reads like a description of what no longer exists, perhaps in the commit message but likely not needed here.

internal/server/loopback.go
```diff
@@ -0,0 +1,208 @@
...
+func (c *loopbackCheck) Interval() time.Duration { return c.interval }
+
+// Status runs one exchange and maps its outcome. Any answer, including a
+// rejection, means the chain ran end to end and reports SERVING: that is what
+// keeps the check free of credentials, since it never needs to authenticate,
+// only to be answered. A deadline means the chain did not answer, and a listener
+// that is not accepting or a failure carrying no status at all means the server
+// is not answering either; both report NOT_SERVING.
+//
+// No hysteresis is applied here. A probe's failureThreshold already debounces,
```
pseudomuto, line 108: TIL...hysteresis

pseudomuto (approved):

PR #173, opened 2026-09-16
[ext]: Seal DEKs with a standard set of ciphers

pseudomuto:
KeyMaterial gave extension servers a framing so they stopped hand-rolling binary layouts, but it left the cryptography with them. Writing a provider still meant building an AEAD, generating a nonce, choosing additional data, and re-checking the nonce length on the way back in. Those are the parts where mistakes are silent rather than loud.

Add a key wrapper that takes a lookup for wrapping keys and does the rest. It seals with an AEAD, frames the result as KeyMaterial, and opens it again. AES-256-GCM is the default, but ChaCha20-Poly1305 and XChaCha20-Poly1305 are also available. A caller can register others via `ext.WithCipherFunc`.

Where key bytes come from stays with the implementor. Owning the derivation would mean owning its migration story too (nope), so the example keeps its HKDF and loses everything else.


examples/kms/README.md
```diff
@@ -211,68 +211,82 @@ of 100.
...
-The whole provider is the `server` command you ran in terminal 2, and almost all of it is one file. The crypto lives in
-`server/keyring.go`; the gRPC surface, the bearer token check, TLS, and graceful shutdown all come from `pkg/ext`.
-`keyring`'s `Wrap` and `Unwrap` are what satisfy `ext.KMS`, and `server/main.go` hands them to `ext.Serve` alongside the
-token check and the certificate. Start with `keyring.go` if you are writing one of these against a real key service: it
-is the part you have to replace.
+The whole provider is the `server` command you ran in terminal 2, and it is one method. `server/keyring.go` derives
+keys, and that is all it does; it contains no cryptography beyond the derivation itself. Everything around it comes from
+`pkg/ext`: `ext.NewKeyWrapper` turns that method into an `ext.KMS`, and `ext.Serve` registers it, checks the bearer
+token, serves TLS, and shuts down on a signal. Start with `keyring.go` if you are writing one of these against a real
+key service, because that method is the part you replace:
```
dipali-bhatt, line 218: Could we qualify this to key services that can return the raw wrapping-key bytes? Some managed key services keep their keys non-exportable and provide only wrap and unwrap APIs. Those implementations would use KeyMaterial for the common framing but continue implementing ext.KMS directly instead of using NewKeyWrapper.

dipali-bhatt (commented):
Thanks for putting this together. For managed key services such as SF KMS, the wrapping key is non-exportable, so the extension cannot return the raw key bytes required by NewKeyWrapper.
Would you consider adding a companion helper, that defines provider-neutral wrap and unwrap callbacks and implements the surrounding ext.KMS and KeyMaterial construction and parsing? The extension callbacks would continue to own authentication, API calls, algorithm and field validation, integrity protection, error mapping, and backend-specific key semantics.
This would keep the solution provider-neutral while allowing managed-KMS integrations to avoid implementing their own durable record encoding and decoding.

Vaughan-Temporal (approved):

pseudomuto:
> Would you consider adding a companion helper, that defines provider-neutral wrap and unwrap callbacks and implements the surrounding ext.KMS and KeyMaterial construction and parsing?

Absolutely, that makes sense to me. I'll think about this and post an update here shortly.

pseudomuto:
Hey @dipali-bhatt, I added another KMS implementation called KeySealer. It's similar to KeyWrapper, but it never gets a key back to handle encryption; it delegates that elsewhere. You can see the implementation in #177

PR #176, opened 2026-09-18
[codecserver]: Serve the Temporal codec server endpoints

pseudomuto:
Payloads the proxy seals are unreadable to any client that reaches a Temporal Service without passing through the gateway. That is the Temporal Cloud UI, which calls a codec endpoint from the user's browser, and `temporal --codec-endpoint ...`. A UI/CLI pointed at the proxy already renders cleartext, since inbound opening runs whether or not sealing is enabled, so Temporal Cloud UI and the CLI using a codec endpoint are the intended cases for the codec server.

Add internal/codecserver, serving /encode, /decode and /download (POST) on its own HTTP port. Each of these is also available under a Namespace path segment because the CLI can substitute the Namespace into the endpoint URL while the UI sends it as a header (X-Namespace). It applies the same payload codec chain the per-upstream proxies do.

Both clients (Temporal Cloud UI and Temporal CLI) name Namespaces the way the upstream does (translated names), while the _seal_ path keys its key policy on the pre-translation local name. Resolve that with a forward map built at config load, computing each policy Namespace's remote name through that upstream's own rules. It is exact and honours explicit local/remote mappings, where inverting the translation would be lossy (e.g. prefix/suffix trim).


pseudomuto:
@Vaughan-Temporal, I know this is a large diff (though lots of docstrings, markdown, and tests). I figured breaking this into multiple PRs wouldn't help since it's either a codec server or not; however, if this is too large to review, let me know and I'll split it up.

